PR #88, opened 2014-05-25
afc: new API using sendfile() for fast and light operation

kamstrup:
Adds one new public function:

  afc_file_write_from_fd(): Wraps sendfile() on Unix
    to avoid copying data between kernel and user space when
    transfering regular files. Falls back to regular read/write
    and non-Unices

Using sendfile() might only be a smidgeon faster than hand optimized
regular read/write ops, but it puts a noticeably lighter load on the
system because it avoids memory churn and context switches. For
non-optimized read/write loads sendfile is about 8 times faster.

Also - the API is nice to use :-) What was 10-20 lines of tricky
buffer juggling is now a 1-liner for most situations.

I tested all #ifdef chunks extensively by transferring files back and forth and checking md5 sums. Everything is working. I have branches porting ideviceinstaller and afcclient to use sendfile if we land this.

I do not have a BSD setup to test on so *BSD falls back to read/write like windows will. Enterprising souls should easily be able to add native FreeBSD sendfile support (there is a TODO in the code about it).

### DIFF
--- a/include/libimobiledevice/afc.h
+++ b/include/libimobiledevice/afc.h
@@ -235,6 +235,24 @@ afc_error_t afc_file_read(afc_client_t client, uint64_t handle, char *data, uint
 afc_error_t afc_file_write(afc_client_t client, uint64_t handle, const char *data, uint32_t length, uint32_t *bytes_written);
 
 /**
+ * Write data from a file directly to and AFC file handle without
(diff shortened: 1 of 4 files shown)


nikias:
While I really liked this approach I think it's finally time to close this, since AFC now also uses SSL on USB connections.

PR #160, opened 2014-12-13
cython: Add receive/receive_timeout methods for iDeviceConnection

upbit:
Tested on iOS7.1.2/Python2.7:

~~~python
#!/usr/bin/env python

import sys
import plist
from imobiledevice import *

# modify from https://github.com/libimobiledevice/libimobiledevice/blob/master/dev/filerelaytest.c

def lockdown_get_service_client(service_class):
	ld = LockdownClient(iDevice())
	return ld.get_service_client(service_class)

def file_relay_get_UserDatabases():
	filerelay = lockdown_get_service_client(FileRelayClient)
	conn = filerelay.request_sources(["UserDatabases"])

	sys.stdout.write("Write dump.cpio.gz:\n")
	with open("dump.cpio.gz", "wb+") as dumpfile:
		while (1):
			try:
				#data = conn.receive_timeout(4096, 10)
				data = conn.receive(4096)
				dumpfile.write(data)

				sys.stdout.write(".")
			except Exception, e:
				if (e.code == -2):		# no more data
					break
				raise e

	print "Done"

def main():
	file_relay_get_UserDatabases()

if __name__ == '__main__':
	main()
~~~

### DIFF
--- a/cython/file_relay.pxi
+++ b/cython/file_relay.pxi
@@ -11,6 +11,7 @@ cdef extern from "libimobiledevice/file_relay.h":
         FILE_RELAY_E_MUX_ERROR = -3
         FILE_RELAY_E_INVALID_SOURCE = -4
         FILE_RELAY_E_STAGING_EMPTY = -5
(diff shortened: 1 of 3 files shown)


FunkyM:
Thanks, pushed to git master (slightly modified).

PR #171, opened 2015-01-16
Win32 support for idevicecrashreport

qmfrederik:
This patch makes idevicecrashreport build on Windows using MinGW.


nikias:
Pushed to master with slight differences.

PR #176, opened 2015-01-25
idevicesyslog: Add -f/--flush argument for automatic stdout flushing

alopatindev:
Sometimes there's just no API to execute fflush(stdout) for external programs.
I ran into such problem while was using QProcess from Qt framework.

http://comments.gmane.org/gmane.comp.lib.qt.general/44548


nikias:
Made idevicesyslog flush by default when encountering a line break.

PR #219, opened 2015-06-23
configure.ac: Don't always explicitly disable openssl

somasis:
This changes configure.ac to use a more correct usage of `AC_ARG_ENABLE`.
Right now openssl is disabled if you pass any form of` --enable-openssl`, `--disable-openssl`,
etc. This fixes that and only disables openssl if you pass `--disable-openssl`.


nikias:
Merged after rebase. Thanks for reporting!

PR #237, opened 2015-09-24
Changes for upstream

zbalaton:
This set contains some small fixes and adding a new function to get the file descriptor of an idevice connection which will be needed for the improved fdr implementation in idevicerestore. Please pull or comment.


xdeng:
✔

PR #245, opened 2015-10-14
configure: pthreads is not required under win32

artygus:


nikias:
Merged after rebase with rephrases commit message.

PR #250, opened 2015-10-21
Fix installation_proxy when using GnuTLS instead of OpenSSL

saurik:
The core problem behind libimobiledevice/ideviceinstaller#11 (which was closed without being fixed) is solved with the following patch, which brings the GnuTLS code to match the behavior of the OpenSSL logic (doing so by minimizing the part controlled by preprocessor conditionals).


nikias:
Thanks Jay. Merged it (after rebasing).

PR #265, opened 2015-12-09
Update manpages with options from --help.

pteromys:
For idevicebackup2 and ideviceinfo.

I kept forgetting that `idevicebackup2 unback` was a thing I could do.

Differences highlighted using https://gist.github.com/pteromys/8e45072041afeb2922c0

Aside: I noticed that in 96101a1231a4ddfeb40fd738a24e108a3a904048, ideviceinfo's "the first connected device" became "a connected device" in `--help`. Maybe that ought to be updated in the manpage too, but I left it alone so that this commit could stay in the realm of mere copy+paste.

### DIFF
--- a/docs/idevicebackup2.1
+++ b/docs/idevicebackup2.1
@@ -31,6 +31,9 @@ prints usage information.
 .B backup
 create backup for the device.
 .TP
+.B \t\-\-full
+force full backup from device.
+.TP
 .B restore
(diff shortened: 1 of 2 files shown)


mexmer:
unback is not a thing anymore i think since io 10 or 11

nikias:
Resolved with commit 2ae528a.

PR #296, opened 2016-03-21
Use TLSv1_method. works with newer openssl

dweinstein:
-  openssl is returning NULL for `SSLv3_method()` on some distros

### DIFF
--- a/src/idevice.c
+++ b/src/idevice.c
@@ -678,7 +678,7 @@ LIBIMOBILEDEVICE_API idevice_error_t idevice_connection_enable_ssl(idevice_conne
 	}
 	BIO_set_fd(ssl_bio, (int)(long)connection->data, BIO_NOCLOSE);
 
-	SSL_CTX *ssl_ctx = SSL_CTX_new(SSLv3_method());
+	SSL_CTX *ssl_ctx = SSL_CTX_new(TLSv1_method());
 	if (ssl_ctx == NULL) {
 		debug_info("ERROR: Could not create SSL context.");
 		BIO_free(ssl_bio);


mexmer:
this will not work with elder ioses, because they use sslv3
if anything, you should use TLS_method() which has SSLv3 fallback
see
https://www.openssl.org/docs/manmaster/ssl/SSL_CTX_new.html

dweinstein:
```
$ grep "TLS_method" /usr/include/openssl/*.h
/usr/include/openssl/ssl.h:const SSL_METHOD *DTLS_method(void); /* DTLS 1.0 and 1.2 */

```

I wasn't able to find the `TLS_method` defined. I'm on `openssl 1.0.2.g-3`.

dweinstein:
also it looks like at the link you sent TLS_method is just a redefine of `SSLv23_method` the archlinux build of libimobiledevice uses `SSLv23_method` and it fails to do calls like `ideviceinfo` with it.

mexmer:
well that page refers to last revision of openssl. since you have not specified version of openssl you use, was obvious to link latest.

regardless on windows i use SSLv23_method with openssl 1.0.1k without problems even with new iOS 9.3 (runnin on iphone 6), i suspect something is wrong with your setup.

dweinstein:
does `ideviceinstaller -l` work for you linked to your imobiledevice?

mexmer:
i'm not sure, if it can be compiled under windows (have not tested it), i have own tools using libimobile and those work. libimobile is built from 1.2.0 source on github.

nikias:
My latest commit changes it to use SSLv23_method() which should be available from my understanding. In OpenSSL >= 1.1.0 this is actually an alias vor TLS_method().

dweinstein:
@nikias did you try doing `ideviceinfo` w/ your device w/ `SSLv23_method` and that worked?

nikias:
works for me, tested on macOS El Capitan and Ubuntu 15.04

nikias:
Ok I just tested with Ubuntu 16.04 and it neither works before and after my change. That stupid SSLv3 disable patch is really killing it...

nikias:
actually, I was wrong and it does work. it failed because it picked up an older libimobiledevice version that was installed. So I get:
```
$ tools/ideviceinfo -k ProductVersion
10.0
```
Also ideviceinstaller works if you make sure that it links against the self-built libimobiledevice.

nikias:
Turned out older devices (iOS 8) didn't work anymore on El Capitan with SSLv23_method(). TLSv1_method() seems to work fine.

nikias:
Merged to master. Seems like the best solution. Thanks for reporting!

PR #303, opened 2016-04-05
Define a `diagnostics_relay_action_t` enum

qmfrederik:
Replace a bunch of constants which can be passed as flags with a single `diagnostics_relay_action_t` enum.

Amongst others, makes it easier to generate a .NET front-end for libimobiledevice.


qmfrederik:
@nikias , I think this PR is still relevant (most other libimobiledevice modules use an enum). What do you think?

nikias:
Merged after rebase, followed by a commit to update the cython interface.

PR #308, opened 2016-04-19
idevicedebug: show error if container info not found

mringwal:
If the container info cannot be retrieved, e.g. because of an invalid bundle id, it crashes setting the working directory as working_dir == NULL.


nikias:
Merged after rebase and fixing of whitespace error :) Thanks Matthias!

PR #335, opened 2016-07-02
Ensure python3 compatibility

bjoernhaeuser:
I tried to run the cython bindings on Python 3.5. There were some errors due to the stricter handling of strings vs bytes.

Therefore default everywhere to bytes.


bjoernhaeuser:
This PR prevents to get this through here: https://github.com/Homebrew/homebrew-core/pull/2643

PR #341, opened 2016-07-12
lockdown: Propagate lower level errors to callers instead of just returning unknown error

zbalaton:
Small patch to allow callers to get a better error code instead of just a generic error to know what went wrong.

### DIFF
--- a/src/lockdown.c
+++ b/src/lockdown.c
@@ -136,6 +136,36 @@ static lockdownd_error_t lockdownd_strtoerr(const char* name)
 }
 
 /**
+ * Convert a property_list_service_error_t value to a lockdownd_error_t
+ * value. Used internally to get correct error codes.
+ *
+ * @param err A property_list_service_error_t error code
+ *
+ * @return A matching lockdownd_error_t error code,
+ *     LOCKDOWND_E_UNKNOWN_ERROR otherwise.
+ */
+static lockdownd_error_t lockdownd_error(property_list_service_error_t err)
+{
+	switch (err) {
+		case PROPERTY_LIST_SERVICE_E_SUCCESS:
+			return LOCKDOWN_E_SUCCESS;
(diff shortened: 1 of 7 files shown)


zbalaton:
I've updated this to current HEAD now and also included some other small fixes. Will this be merged now?

src/lockdown.c
```diff
@@ -722,6 +733,9 @@ LIBIMOBILEDEVICE_API lockdownd_error_t lockdownd_client_new_with_handshake(idevi
 		ret = lockdownd_pair(client_loc, NULL);
 		if (LOCKDOWN_E_SUCCESS == ret) {
 			ret = lockdownd_validate_pair(client_loc, NULL);
+			if (ret == LOCKDOWN_E_MUX_ERROR) {
+				ret = LOCKDOWN_E_INVALID_HOST_ID;
+			}
```
nikias: why does mux error mean invalid host id?

nikias:
Looks good to me, except I was wondering about the if (ret == LOCKDOWN_E_MUX_ERROR) ret = LOCKDOWN_E_INVALID_HOST_ID is the right way to do it.

zbalaton:
LOCKDOWN_E_INVALID_HOST_ID is normally returned when pairing should be tried again. When the trust button is tapped, the device breaks connection which will result in MUX_ERROR but this at this point means that pairing was successful and should be retried so the error is translated to the error code which is normally returned for this case in other places. Using other error code would need clients already checking for INVALID_HOST_ID to be changed to also check for that other error code so this is the right way for compatibility.

zbalaton:
I've removed the disputed hunk converting LOCKDOWN_E_MUX_ERROR to LOCKDOWN_E_INVALID_HOST_ID on pairing failure so the rest could be merged. We can consider this change separately in subsequent commit. Anything else holding this back?

nikias:
No, I will merge it.

PR #342, opened 2016-07-13
Fix idevicebackup2 for iOS10

sapieneptus:
`idevicebackup2` was failing to remove the Snapshots dir because it had some empty dirs inside of it. 

We can resolve this by using `fts` to walk the ‘Snapshots’ and remove
empty subdirs first instead of just calling `remove()`.


nikias:
Fixed with commit https://github.com/libimobiledevice/libimobiledevice/commit/6aecc7228dbb30b4db48374ffb4ae4a1616bdd95

PR #345, opened 2016-07-18
Provide access to idevice_connection for property_list_service

mringwal:
Provide access to idevice_connection for property_list_service

```
Some services first expect a command as a property list, then pure data (e.g. a zip stream)
before sending a result again as a property list. Providing direct access to the
underlying idevice_connection avoids re-implementing the property list service serialization.
```

### DIFF
--- a/include/libimobiledevice/property_list_service.h
+++ b/include/libimobiledevice/property_list_service.h
@@ -162,6 +162,18 @@ property_list_service_error_t property_list_service_enable_ssl(property_list_ser
  */
 property_list_service_error_t property_list_service_disable_ssl(property_list_service_client_t client);
 
+/**
+ * Get underlying idevice_connection of the given property list service client
+ *
+ * @param client The connected property list service client to get idevice_connection
+ * @param connection Pointer to an idevice_connection_t that will be filled
+ *   with the necessary data of the connection.
+ *
+ * @return PROPERTY_LIST_SERVICE_E_SUCCESS on success,
+ *     PROPERTY_LIST_SERVICE_E_INVALID_ARG if client or connection is NULL.
+ */
+property_list_service_error_t property_list_service_get_connection(property_list_service_client_t client, idevice_connection_t * connection);
+
 #ifdef __cplusplus
 }
 #endif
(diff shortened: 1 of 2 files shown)


youngchief-btw (approved):

nikias:
From a hierarchical standpoint I think this should be a function that returns the parent service_client_t of a property_list_service_client_t, and then a helper that returns the idevice_connection_t from a given service_client_t.

mringwal:
@nikias I don't remember anything about this PR (from 5.5 years ago). I'm sure your counter-proposal is better and you should move forward with that :)

mexmer:
this sounds to me like similar situation to housearrest service, where you first send PLIST commands, then connection switches to AFC mode. tbh. not sure what will be best approach here. also i don't remember seeing any service (Except housearrest), that actually switches modes and needs direct connection. 

@mringwal if you can remember what service was that ... maybe there is already implementation of that

mringwal:
@mexmer According to internal commit messages, I was working on Apple Watch support for a client. So, this suggestion might have been relevant to implement support for the company proxy ("com.apple.companion_proxy"). I think it allows to forwards a TCP socket to the Watch. (I'm not involved in iOS

nikias:
Ah ok that makes sense now. House arrest is what also came to my mind but that already has a service abstraction/interface.
For AFC on watches we obviously would need to pass the service_client_t handle, so I guess my suggestion would be best here to allow access to each "superclass" of whatever service type we deal with.

nikias:
See 854ab9a5d9d04fb884459c5092f6620faa99673b.

mringwal:
@nikias  Thanks, looks good!

mexmer:
> See [854ab9a](https://github.com/libimobiledevice/libimobiledevice/commit/854ab9a5d9d04fb884459c5092f6620faa99673b).

guess, i will use this, instead of my hacky workaround for working with companion started services

nikias:
> guess, i will use this, instead of my hacky workaround for working with companion started services

I personally would define the private structs and then do `client->parent->connection` but that's kind of a bad advice I guess 🤣🙈

PR #350, opened 2016-07-27
On newer macOS versions, /var/db/lockdown/ is 700.

saurik:
So, I've been working on tracking down a widely-reported LOCKDOWN_E_INVAILD_HOST_ID error while trying to do essentially anything. It seemed to mostly be affecting users of macOS 10.12, with a few people on OS X 10.11 (if so, probably only the absolute latest versions). Working with @conradev (who is actually running the latest version of macOS), we determined that the issue is that /var/db/lockdown is now mode 700, and so the code in lockdown.c for `userpref_has_pair_record` no longer works. (The only code which relies on that folder being accessible other than that function is `userpref_get_paired_udids`, which is only used by idevicepair as what is more of a low-level debugging tool.) The code in `lockdownd_do_pair` seriously is asking "is the record on disk? ok, let's ask for a copy from usbmuxd", so I simply removed the "is the record on disk?" part, as if it isn't there then we can expect usbmuxd to not return anything (right? ;P).

### DIFF
--- a/common/userpref.c
+++ b/common/userpref.c
@@ -192,35 +192,6 @@ int userpref_read_system_buid(char **system_buid)
 }
 
 /**
- * Determines whether this device has been connected to this system before.
- *
- * @param udid The device UDID as given by the device.
- *
- * @return 1 if the device has been connected previously to this configuration
- *         or 0 otherwise.
- */
-int userpref_has_pair_record(const char *udid)
-{
-	int ret = 0;
-	const char *config_path = NULL;
-	char *config_file = NULL;
-	struct stat st;
-
-	if (!udid) return 0;
-
-	/* first get config file */
-	config_path = userpref_get_config_dir();
-	config_file = string_concat(config_path, DIR_SEP_S, udid, USERPREF_CONFIG_EXTENSION, NULL);
-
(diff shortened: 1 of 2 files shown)


nikias:
@saurik Thanks for spotting this. This happened during the transition from the direct pair record handling to usbmuxd handling the records :)

I'd like to see the commit split into two: one for the lockdown.c change and one for the userpref.c change.

Please adapt the lockdown commit message to something like:
```
lockdown: remove unnecessary check for pair record file during pairing

During device pairing the code in lockdownd_do_pair() is checking if there is a pair record on disk for the current device, and then requests it from usbmuxd. This additional check is not only unnecessary since usbmuxd can only return a pair record if it exists, but is also causing issues on newer versions of macOS where /var/db/lockdown is mode 700.
```

And the userpref one should be something like:
```
userpref: Remove obsoleted function userpref_has_pair_record()

Since pair records are meanwhile handled by usbmuxd there is no need to check for the existence of a pair record on disk. Asking usbmuxd for a pair record of a given UDID is sufficient to know if it exists or not.
```

Then I will merge it right away.

saurik:
@nikias Done ;P.

PR #352, opened 2016-08-02
idevice: Update GnuTLS code to support iOS 10

saurik:
As of iOS 10 beta 4, the GnuTLS implementation idevice_connection_enable_ssl
needs to be updated to support TLS. Using +VERS-TLS-ALL did not work on some
of the devices I tested and I wasn't sure how to fix it, but +VERS-TLS1.0 is
working on every device I've tested: iOS 9.0.2, 10.0b4, 8.1.1, 6.0, and 3.0.


saurik:
(To be more honest: +VERS-TLS-ALL didn't work on *any* of the devices I tested, but I simply didn't test it on as many ;P.)

nikias:
Thanks Jay!

PR #362, opened 2016-08-26
Removing snapshot directories which failed during idevicebackup2 on iOS10

sohgoh:
I tried to backup iOS 10 device data by idevicebackup2 but failed. The reason is snapshot directory  has any empty directories in itself. Directory structure of snapshot fetched from iOS is changed from iOS10.

iOS9 or earlier: backup_dir/{UDID}/backup_files
iOS10: backup_dir/{UDID}/([0-9a-f]{2})/backup_files

This fix works for the same purpose as #342 but by another approach. It works to remove empty directory "00-ff" only in snapshot directory on iOS 10.

### DIFF
--- a/tools/idevicebackup2.c
+++ b/tools/idevicebackup2.c
@@ -1112,6 +1112,26 @@ static void clean_exit(int sig)
 	quit_flag++;
 }
 
+static void clean_snapshot(char *snapshot_dir) {
+	char subdir[8];
+	uint32_t i = 0;
+	struct stat st;
+
+	for (i = 0; i < 256; i++) {
+		sprintf(subdir, "%02x", i);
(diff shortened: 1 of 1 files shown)


gitsop01:
It would be better and easier to understand the code commit if you actually describe what the commit is fixing such as

"Removing snapshot directories which failed during idevicebackup2 on iOS10."
than
generic message "idevicebackup2 support iOS10"

sohgoh:
Thanks your comment, I changed pull request title.

nikias:
Fixed with commit https://github.com/libimobiledevice/libimobiledevice/commit/6aecc7228dbb30b4db48374ffb4ae4a1616bdd95

PR #426, opened 2016-11-29
Fetching IOKit tree via idevicediagnostics ioreg failing on iOS 9.1, …

rgadbois:
…but works by entry

### DIFF
--- a/tools/idevicediagnostics.c
+++ b/tools/idevicediagnostics.c
@@ -40,7 +40,8 @@ enum cmd_mode {
 	CMD_SHUTDOWN,
 	CMD_DIAGNOSTICS,
 	CMD_MOBILEGESTALT,
-	CMD_IOREGISTRY
+	CMD_IOREGISTRY,
+    CMD_IOREGISTRY_ENTRY
 };
 
 static void print_xml(plist_t node)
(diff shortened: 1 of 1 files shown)


nikias:
@rgadbois Can you redo the pull request with correct indentation using tabs instead of spaces? Thanks.

reidhoch:
Are we still waiting on this to be fixed?

daniel-bluesea:
@nikias @rgadbois can we still retrieve the IODeviceTree or is it obsolete? 
do we have a "new name" for it for the new ioregentry command?

progressnerd:
Did it ever work by plane alone? Also, I believe if you're matching by Class or Name that a Plane is optional, like this API:
  https://developer.apple.com/documentation/installerjs/ioregistry/1812186-matchingclass
  https://developer.apple.com/documentation/installerjs/ioregistry/1812200-matchingname

I'm thinking one ioreg command with flags for class, name, and plane.

nikias:
I merged this with correct indentation and some small cosmetic changes.

progressnerd:
Unfortunately, this change offers no way to query by entry class, like IOPMPowerSource.

PR #430, opened 2016-12-07
Fix for connecting elder ios devices

mexmer:
iOS 5 and iOS 4 devices (tested) have problem with autonegotiation of SSL protocol version when enabling secure connection.

This will force using SSL3 on this devices (actually patch will force SSLv3 on all devices prior TLS1 support - eg. prior iOS 9).

Solution is suboptimal, since it queries iOS version every time you try to enable SSL, optimal will be, if iOS major number is stored in one of structures used, and queried only on device connect.

code is written to support compilation with openssl < 1.1, but also on openssl 1.1 with disabled deprecated apis

### DIFF
--- a/include/libimobiledevice/lockdown.h
+++ b/include/libimobiledevice/lockdown.h
@@ -464,6 +464,17 @@ lockdownd_error_t lockdownd_data_classes_free(char **classes);
  */
 lockdownd_error_t lockdownd_service_descriptor_free(lockdownd_service_descriptor_t service);
 
+/**
+ * Returns the iOS version of the device from lockdownd.
+ *
+ * @param client An initialized lockdownd client.
+ * @param udid Holds the iOS version string of the device. The caller is responsible
+ *  for freeing the memory.
+ *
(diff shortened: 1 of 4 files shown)


mexmer:
@FunkyM if you have any thoughs where to put OSMajor version, instead of this, let me know.

In our software we query osMajor version upon device detecion (connection to computer), and is store in device connection object.

iOS version is one of values you can query regardless device is locked or not.

nikias:
@mexmer I tried latest git HEAD on ubuntu, with GnuTLS and OpenSSL, and both work with an iPhone 4 on iOS 5.0.1 without issues. The code is using TLSv1_method() now. Have you tried it?

nikias:
I tried with OpenSSL 1.0.2g

mexmer:
my version was 1.0.1k have not tested 1.0.2, since i moved to 1.1 from 1.0.1

mexmer:
with 1.0.2 as well as 1.1 it's working without issues. maybe there was problem with autonegotiation in 1.0.1 ... closing PR, since it's not needed.

PR #446, opened 2017-01-02
Improve usage of gnutls library

nmav:


nikias:
Have you checked if this works with older devices/iOS versions?

nmav:
No. However, this change shouldn't affect older devices or iOS versions. TLS1.0 and AES (the ciphers in the previous string) are enabled by the gnutls_set_default_priority().

mexmer:
since autonegotiation with TLS 1.2 enabled breaks compatibility with elder iOS devices, i believe this will do so also. @nmav please retest your solution with iOS device running iOS 5 or lower.

nmav:
Hmmm, in that case I withdraw that part of the patch. Let's handle it when the time comes.

mexmer:
@nmav you might wait for saurik to retest. i compile libimobile and related stuff against openssl, so can't test your stuff without completely rebuilding everything.

nikias:
Merged to master after rebase.

PR #453, opened 2017-01-10
property_list_service: Remove packet length check when receiving plist data

crevetor:
There are services that would send really large plist data, e.g. when listing provisioning profiles. Instead of forcing the data to be less than 16MB we try to allocate a buffer as large as requested. If the allocation fails the function returns with an error.

### DIFF
--- a/src/property_list_service.c
+++ b/src/property_list_service.c
@@ -193,59 +193,56 @@ static property_list_service_error_t internal_plist_receive_timeout(property_lis
 		debug_info("initial read failed!");
 		return PROPERTY_LIST_SERVICE_E_MUX_ERROR;
 	} else {
-		pktlen = be32toh(pktlen);
-		if (pktlen < (1 << 24)) { /* prevent huge buffers */
-			uint32_t curlen = 0;
-			char *content = NULL;
-			debug_info("%d bytes following", pktlen);
-			content = (char*)malloc(pktlen);
-			if (!content) {
-				debug_info("out of memory when allocating %d bytes", pktlen);
-				return PROPERTY_LIST_SERVICE_E_UNKNOWN_ERROR;
-			}
+		uint32_t curlen = 0;
+		char *content = NULL;
 
-			while (curlen < pktlen) {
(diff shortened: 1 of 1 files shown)


nikias:
@crevetor can you update the pull request and change the commit message to
```
property_list_service: Remove packet length check when receiving plist data

There are services that would send really large plist data, e.g. when listing provisioning profiles. Instead of forcing the data to be less than 16MB we try to allocate a buffer as large as requested. If the allocation fails the function returns with an error.
```

mexmer:
@crevetor do you have any aditional info, which services can send larger plist (either binary or xml) than 16MB?
so far, most service had browse mode (like installation proxy), or different mechanism to send large plist in chubks

nikias:
I personally haven't observed anything that large myself. The provisioning profile service (misagent) doesn't seem to expect these large amounts of data, otherwise they would have a more appropriate API to handle that. Maybe
But anyway I think it can't hurt to remove the limitation of 16MB which was arbitrarily chosen; also for sending plists to the device we don't check the size at all, which could easily fail as well whenever we run out of memory. For example, when restoring a device, baseband data can be around 30mb and is also sent as a single plist. The restored service doesn't send anything large from the device though.

crevetor:
@mexmer I personnaly experienced this limit while retrieving provisioning profiles from phones. We have phones with around 70 provisioning profiles weighing 2MB each which amounts to 140MB to pull down from the phone. If you want I can send you the debug logs of downloading those PP from the phone.

mexmer:
@crevetor no need for logs, just wanted to know, where you experienced it.

i for one had long time limit of 2MB buffer in my solution for any service response, until i got some listing from AFC (i think it was photo folder), when they pushed over 8 MB, then i found out libimobile uses 16MB as "limit" for invalid packet, so i increased limit to 16MB and no issues since then.

but if there are services, that push more, than 16MB, i was just curious if it's common, or just rare for only some services.

mexmer:
just found this issue referenced in our source code, we still did have hardlimit for packet size check, but upped to 256MB, today i got phone, that sent 364MB long packet ... guess i will remove this check from our sourcecode too.

PR #480, opened 2017-03-20
Add idevicelocation tool

robbertvanginkel:
Set device location for specified lat/long on an attached device.

Usage:
```
  $ idevicelocation <lat> <long>
  $ idevicelocation --stop
```
Uses `com.apple.dt.simulatelocation`/`DTSimulateLocation` from the
developer disk image under the hood.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -47,6 +47,7 @@ tools/idevicedate
 tools/idevicedebugserverproxy
 tools/idevicediagnostics
 tools/ideviceenterrecovery
+tools/idevicelocation
 tools/idevicename
 tools/idevicepair
 tools/ideviceprovision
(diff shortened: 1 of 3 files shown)


nudge:
This is amazing. Fantastic work Robbert!

Dantee296:
Hi can you tell me how to enable GPS (Location Services) remotely 
can we use this tool to enable it ?

nghiadhd:
@Dantee296 
I must run XCUITest script to turn on LocationService before do simulate location

nghiadhd:
@robbertvanginkel 
Seem 'idevicelocation --stop' doesn't work. It doesn't switch back to GPS location

Dantee296:
@nghiadhd  u mean automation using appium

nghiadhd:
@Dantee296 
You can use Appium or just write a XCUITest script and use xcodebuild test to run this script

daniel-bluesea:
what is the format of <lat> <long>?
what is the impact on the attached device?

umutuzgur:
Awesome PR! Any reason why this is not being merged?

c00ni:
@nghiadhd It does reset location, it just might take a minute.

nghiadhd:
@truonggiang0710 
I don't check location status, I just get location by my MAC and set it into the device before actually call stopping
That is fast way than stop

truonggiang0710:
Hi @nghiadhd , thanks for your reply. Do we have any way to enable/disable location service in Setting on iPhone ?

nikias:
Hi, I had this lying around on my disk since some time back in 2016. I just pushed it to git. I called it `idevicesetlocation` and the usage is slightly different.

PR #484, opened 2017-03-28
Fix segmentation fault occuring when installing app by ideviceinstaller

sohgoh:
Fix segmentation fault occuring when installing application by ideviceinstaller on iOS 10.3 or later.
iOS 10.3 or later device returns CFBundleIdentifier response between installation status responses.

ideviceinstaller issue: https://github.com/libimobiledevice/ideviceinstaller/issues/70

### DIFF
--- a/include/libimobiledevice/installation_proxy.h
+++ b/include/libimobiledevice/installation_proxy.h
@@ -497,6 +497,14 @@ void instproxy_client_options_free(plist_t client_options);
  */
 instproxy_error_t instproxy_client_get_path_for_bundle_identifier(instproxy_client_t client, const char* bundle_id, char** path);
 
+/**
+ * Gets CFBundleIdentifier of a status.
+ *
+ * @param status The dictionary CFBundleIdentifier response to use.
+ * @param name Pointer to store the name of bundle identifier.
+ */
+void instproxy_get_cfbundle_identifier(plist_t status, char **name);
+
 #ifdef __cplusplus
 }
 #endif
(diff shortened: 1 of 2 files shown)


sohgoh:
Sorry I found bug this fix, close  it.

iOSShohrab:
can you please give me ideviceinstaller.exe with this fix. just want to run it into my machine

iOSShohrab:
Trying to build using this fix for win32 but not working. Any idea?

Jun-GitHub:
may be this bug is not fix.  I still meet this problem on 

> macOS 10.12.4

cthinkel:
We were having the issues described here: https://github.com/libimobiledevice/ideviceinstaller/issues/70

After making these changes, the issue was resolved. @sohgoh Can you elaborate as to why you closed this? Did these changes introduce another, unintended issue?

cthinkel:
Nvm, we traced it down to a problem in the ideviceinstaller code: https://github.com/libimobiledevice/ideviceinstaller/issues/70#issuecomment-293712807

sohgoh:
@cthinkel 
I closed this  because there was a problem in the case where the installation fails as in the case of an ipa file with an incorrect provisioning profile specified. After fixing the problem, it is working in my environment(OS X 10.12.4 & iOS 10.3.1 & iPhone 6/6s/SE/7/7+). However, I checked in the OSX environment only, I do not know whether it will work in other environments(windows, linux).

https://github.com/sohgoh/libimobiledevice/commits/master

PR #494, opened 2017-04-25
Improve OpenSSL support

cfergeau:
First commit fixes a double-free, second and 3rd one some minor compile-time warnings.

### DIFF
--- a/src/idevice.c
+++ b/src/idevice.c
@@ -51,6 +51,14 @@
 #include "common/debug.h"
 
 #ifdef HAVE_OPENSSL
+
+#if OPENSSL_VERSION_NUMBER < 0x10002000L
+static void SSL_COMP_free_compression_methods(void)
+{
+	sk_SSL_COMP_free(SSL_COMP_get_compression_methods());
+}
+#endif
+
 static mutex_t *mutex_buf = NULL;
 static void locking_function(int mode, int n, const char* file, int line)
 {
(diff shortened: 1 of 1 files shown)


nikias:
Can you change the first commit to:

```
#if OPENSSL_VERSION_NUMBER < 0x10002000L
	sk_SSL_COMP_free(SSL_COMP_get_compression_methods());
#else
	SSL_COMP_free_compression_methods();
#endif
```
Because ```SSL_COMP_free_compression_methods()``` was introduced in 1.0.2 but we want to support down to OpenSSL 0.9.8.

src/idevice.c
```diff
@@ -37,10 +37,15 @@
...
+#if OPENSSL_VERSION_NUMBER < 0x10100000
 #if OPENSSL_VERSION_NUMBER >= 0x10000001L
-/* since OpenSSL 1.0.0-beta1 */
+/* since OpenSSL 1.0.0-beta1, but deprecated in OpenSSL 1.1.0 */
 #define HAVE_ERR_REMOVE_THREAD_STATE 1
-#endif
+#else
+#define HAVE_ERR_REMOVE_STATE 1
+#endif /* OPENSSL_VERSION_NUMBER */
+#endif /* HAVE_OPENSSL */
```
nikias: the comment is wrong, this ```#endif``` is for ```/* OPENSSL_VERSION_NUMBER < 0x10100000L */```

src/idevice.c
```diff
@@ -55,7 +55,7 @@
 #include "common/thread.h"
 #include "common/debug.h"
 
-#ifdef HAVE_OPENSSL
```
nikias: Just for completeness, can we keep the additional ```#ifdef HAVE_OPENSSL``` around this?

PR #501, opened 2017-05-13
check return value for "asprintf", eliminate compiler warnnings

chenbd:
this commit eliminate compiler warnings about asprintf, the check the return value for asprintf.

### DIFF
--- a/src/debugserver.c
+++ b/src/debugserver.c
@@ -561,14 +561,16 @@ LIBIMOBILEDEVICE_API debugserver_error_t debugserver_client_set_argv(debugserver
 
 	debugserver_error_t result = DEBUGSERVER_E_UNKNOWN_ERROR;
 	char *pkt = NULL;
-	int pkt_len = 0;
+	size_t pkt_len = 0;
 	int i = 0;
 
 	/* calculate total length */
 	while (i < argc && argv && argv[i]) {
 		char *prefix = NULL;
-		asprintf(&prefix, ",%d,%d,", (int)strlen(argv[i]) * 2, i);
-		pkt_len += (int)strlen(prefix) + (int)strlen(argv[i]) * 2;
+		int ret;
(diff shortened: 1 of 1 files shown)


Tatsh:
You should not cast the return value of `strlen()` for the `asprintf()` calls. The return value type from `strlen()` is `size_t` which is not `int`, it is `unsigned int` (see `man stddef.h`). The other lines cast because the original data type is `int` and to get rid of warnings (arguably the data type for `pkt_len` should be `unsigned int` so the casts can be removed).

The warning you are getting from the formatting functions like `asprintf()` without the cast is telling you to use `%u`, not to cast the value.

```c
void f() {
    ret = asprintf(&prefix, ",%u,%u,", strlen(argv[i]) * 2, i);
}
```

chenbd:
yes, The return value type from strlen() is size_t. but size_t is unsigned int only on 32 bit systems in my opinion. 

so on my 64bit system the warning is:
debugserver.c:571:3: warning: format '%u' expects argument of type 'unsigned int', but argument 3 has type 'size_t' [-Wformat=]
   ret = asprintf(&prefix, ",%u,%d,", strlen(argv[i]) * 2, i);

Tatsh:
Yes the width is different on 32-bit vs 64-bit. It should be 32-bit on 32-bit systems and 64-bit on 64-bit systems, although this is not a guarantee either. So I am changing my mind and saying to cast in this case, because having an `#ifdef` for 64-bit vs 32-bit is lame when it's only a call to `asprintf` and the return value from `strlen` will not exceed the limit of `unsigned int` on 32-bit or 64-bit. I wish there was a better way without casting and not using preprocessor macros.

neheb:
%zu ?

nikias:
Merged with 94ed8c464dee51191f16b5758f090e920fb4d3e1.

Tatsh:
Even if 5 years later, great.

nikias:
@Tatsh hahahahaha :) sorry took some time to review it 🤣

PR #591, opened 2017-11-22
idevicecrashreport: added option to skip WiFiManager logs

hkivela:


achandr9:
Hello,

Is there a plan to merge this PR with the master branch?
On some iOS devices, I see a lot of WiFiManager crash logs.  It will be useful to have an option that would ignore them.

LaikaN57:
@hkivela @achandr9 @szprutamich I think #822 is a more versatile solution. Consider upvoting that PR.

nikias:
I think the filtering solution #822 is probably more useful. If you think it's still relevant, I suggest to make a PR like #822 but with in inverse filter of matching filenames to skip.

PR #645, opened 2018-03-27
Fix inconsistent declaration parameter name

progressnerd:
See: http://clang.llvm.org/extra/clang-tidy/checks/readability-inconsistent-declaration-parameter-name.html


nikias:
Fixed with commit 8049f1a.

PR #647, opened 2018-03-27
Fix idevicedebug argv encoding

pachoo:
The encoding of the "A" command in idevicedebug had issues when encoding a character with a value > 0x7F such as some UTF-8 characters.


nikias:
I don't think we need the assert. The macros are doing `& 0xf` so the value will always be `x >= 0 && x < 16`.

nikias:
Merged after rebase and without assert.

PR #650, opened 2018-03-30
Update debug.c to consistently output to stderr

progressnerd:
This makes debug_print_line consistent with debug_buffer and among other things, ensures output from `idevicedebug run` can be easily divided into output from the app itself (stdout) from debug output from libimobiledevice (stderr).


nikias:
Merged.

PR #651, opened 2018-03-30
Gracefully fail on invalid bundle ids and fix memory buffer overrun

greg-dennis:
Only decode the response if it's not OK, otherwise the O in "OK" can be interpreted as a hexadecimal and cause a buffer overrun.


nikias:
Merged with f994ee4b849587a9623ea08af064afcc5949d5c8 and cf343da8681b2b8f42377b60ddf09e1bab51be71.

nikias:
Merged with f994ee4b849587a9623ea08af064afcc5949d5c8 and cf343da8681b2b8f42377b60ddf09e1bab51be71.

PR #653, opened 2018-03-31
Better filename selection for idevicescreenshot

progressnerd:
Make it able to use a filename prefix specified by the user and ensure the saved filename is unique on disk.


progressnerd:
Closing this because I think it should use a user-provided file extension if one is given.

PR #654, opened 2018-04-02
Make idevicescreenshot choose a better filename

progressnerd:
If a filename is provided without an extension, it should append the appropriate extension, and it should ensure the saved filename it unique on disk.


youngchief-btw (approved):

nikias:
Merged after rebase.

PR #664, opened 2018-05-10
tools,common: support any port in idevicedebugserverproxy

eliasnaur:
To eliminate crosstalk between multiple proxies and their clients,
add support for binding to any free port provided by the OS to
idevicedebugserverproxy. To bind to any port, leave out the port
argument to idevicedebugserverproxy. In that case, the proxy will
print out a line with the port so clients can connect to it.

This is useful for a CI macOS host with multiple iDevices
connected, and where many independent tests each want their own
proxy instance connected to a particular device.

### DIFF
--- a/common/socket.c
+++ b/common/socket.c
@@ -390,3 +390,24 @@ int socket_send(int fd, void *data, size_t length)
 {
 	return send(fd, data, length, 0);
 }
+
+int socket_get_socket_port(int fd, uint16_t *port)
+{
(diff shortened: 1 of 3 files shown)


nikias:
Merged with b83bc867e61333b61d2f3366e6daa3f6f988074f. Requires https://github.com/libimobiledevice/libimobiledevice-glue/commit/5950ea6d4139394c38c7bdc445725d04113c1fee.

Issue #527, opened 2017-07-17
App backup/restore not working


njarun:
Tried backup with **idevicebackup2.c** and the app data folders are backed up. But the itunes installed application files are not backed up or restored.

Checked with unback, no application file is found in the bundle.

mexmer:
applications (itself) are not part of itunes backup, and never were, only their data.

there is another issue already discussing posibility of restoring applications (or rather, installed application list, so phone will download them from store), but it's not implemented yet.

Issue #343, opened 2016-07-14
Applications dictionary missing in info.plist during mobilebackup2 backup session


NitzanDavari:
I noticed mobilebackup_factory_info_plist_new doesn't create Applications dictionary as iTunes does.
As result the user's apps aren't downloaded automatically in the restored device. 
I noticed the dictionary fills each app with the following information:
app id , ApplicationSINF , PlaceholderIcon , iTunesMetadata.
I do know how to retireve app ids using installation proxy, however the i failed to retrieve the other three ( attempted to use instproxy_lookup but failed - any ideas solution for this

mexmer:
ApplicationSINF and iTunesMetadata can be retrieved from installation proxy, as for PlaceholderIcon, i don't think i have seen that attribute there.

NitzanDavari:
@mexmer  thank you for your reply.
first question, is how do I retrieve ApplicationSINF and iTunesMetadata can be retrieved from installation proxy - i tried with iDeviceInstaller, and i couldn't trace that information - here is the output I received 
[out.txt](https://github.com/libimobiledevice/libimobiledevice/files/367688/out.txt)

And regarding to your second comment - it is related to the springboard - according to my testings, if this dictionary is available in the info.plist the restored device loads with the source device apps like this:
![img_1382](https://cloud.githubusercontent.com/assets/7837708/16899024/9bcf07f8-4bfc-11e6-9f6c-94f64e5fda87.PNG)
and then download the apps to the restored device. When that dictionary isn't available no buffered\placeholders apps are seen in the target device.

Once again thank you for your support!

mexmer:
as for 1st question, i'm not sure, if iDeviceInstaller can or cannot do that, i don't use it. i have my own code, that calls installation_proxy service.
you can specify, which attributes you want in proxy listing output.

as for placeholder icons, if you don't specify them, you just see "generic" download icon, so imo it's not big issue, otherwise you will need to use springboard service, to retrieve app icon, and put it inside that block. although on iOS 9 devices might be possible to retrieve placeholder icon from install_proxy, i have not tested that.

NitzanDavari:
Many thanks, i'll give it a try.
Your assistance is highly appreciated :)

mexmer:
just checked some of my old backups, and Applications dictionary is not there (and i mean backups created by iTunes, not idevicebackup/idevicebackup2).

Most likely this was added by recent version of iTunes, since iTunes versions found i my backups were
[...]

PR #668, opened 2018-05-15
Restore applications when restoring a backup

samuelsadok:
Addresses #527 based on the findings layed out in #343.
Basically this PR makes `imobiledevicebackup restore` create the file `/iTunesRestore/RestoreApplications.plist` with the contents of the `Applications` dictionary of Info.plist.

This does not adhere to the exact sequence that iTunes carries out as described in https://github.com/libimobiledevice/libimobiledevice/issues/343#issuecomment-251716565, but it works which is the most important part for me as an end user.

### DIFF
--- a/tools/idevicebackup2.c
+++ b/tools/idevicebackup2.c
@@ -1343,6 +1343,10 @@ int main(int argc, char *argv[])
 	plist_t info_plist = NULL;
 	plist_t opts = NULL;
 	mobilebackup2_error_t err;
+	uint64_t restore_applications_file = 0;
+	plist_t applications_plist = NULL;
+	char * applications_plist_xml = NULL;
+	uint32_t applications_plist_xml_length = 0;
 
 	/* we need to exit cleanly on running backups and restores or we cause havok */
 	signal(SIGINT, clean_exit);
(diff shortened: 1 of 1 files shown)


AleXoundOS:
I confirm that this works. By using the command `idevicebackup2 --source <device_id> --interactive restore --system --reboot --copy --settings <backup_dir>` after a full backup the device restored all applications (automatically downloaded and installed).

Backup source: iPad 4, iOS: 10.3.3
Restore target: iPad 4 (same)

kalehv:
Any chance of this getting merged?

sribavas:
Does this work on Windows?

samuelsadok:
I would assume so, it doesn't do anything platform specific on the backup host.

tools/idevicebackup2.c
```diff
@@ -1866,6 +1870,51 @@ int main(int argc, char *argv[])
 			}
 			PRINT_VERBOSE(1, "Backup password: %s\n", (backup_password == NULL ? "No":"Yes"));
 
+			/* Write /iTunesRestore/RestoreApplications.plist so that the device will start
+			* restoring applications once the rest of the restore process is finished */
+			applications_plist = plist_dict_get_item(info_plist, "Applications");
+			if (applications_plist) {
+				plist_to_xml(applications_plist, &applications_plist_xml, &applications_plist_xml_length);
+				plist_free(applications_plist);
```
nikias: the plist_free() here is incorrect. it would remove the Applications entry from the info_plist plist_t object.

nikias:
I pushed your changes and added a follow-up commit to move the code into a helper function and also added a command line switch `--skip-apps` to make it not write the file. See 67e3373d2e4bd55795e62b2d51ab99aeddf2406c.

PR #671, opened 2018-05-22
Fix for Python 3

Tatsh:
The Cython module will build against Python 3, but then it won't load because `PyString_AsString` was removed. `PyBytes_AS_STRING` (variant that does not raise exception on NULL) exists in both Python 2.7 and 3.


nikias:
Should be fixed in master via PR #685.

PR #703, opened 2018-09-25
Accommodate 25 digit UDIDs in 2018 iPhones

HelveticaScenario:


O-kasso:
Beat me to it! My alternate fix can be found at #705

djkmiles:
Why isn't this accepted yet? We are depending on this for new Flutter launches.

Issue #702, opened 2018-09-21
idevicescreenshot is broken with iOS 12 iPhone XS Max


umutuzgur:
Hi there,

There is nothing to fix with this issue on the libimobiledevice side because there are crash reports with Screenshotr that comes from the Developer Image. I just wanted to report this so people are aware that there is an issue with the iOS Developer Image release.

kazzmir:
Xcode also has issues getting screenshots with Xs max.

mexmer:
you need to wait, until apple releases new xcode that will work with new iphones. it's similar issue, like with iPhone X, they released iPhone X before xcode, leading to various issues.

PR #705, opened 2018-09-26
Revise device udid validation to account for 2018 iPhones

### LINKED ISSUE #701: UDID format change with iPhone XS
Hi there,

With the release of iPhone XS and XS Max, the iPhone udid format changed it became this value of 25 digits with the hyphen in between. Most of the command line tools have the udid length check which is 40. These are not valid anymore. We should come up with a strategy to handle this case. Maybe removing these checks would be good solution.

Thanks a lot.

O-kasso:
Newer iPhones no longer use the 40-digit hex UDID scheme, instead they use a 25-digit scheme that includes hyphens. Instead of only checking length of provided UDID, check it against a regex that allows for any 25 or 40 digit hex (with hyphens).

Added the new UDID validation function to common utils since it could be useful elsewhere.

This fixes #701 and #702

### DIFF
--- a/common/utils.c
+++ b/common/utils.c
@@ -30,7 +30,7 @@
 #include <sys/time.h>
 #include <inttypes.h>
 #include <ctype.h>
-
+#include <regex.h>
 #include "utils.h"
 
 #ifndef HAVE_STPCPY
@@ -479,5 +479,24 @@ void plist_print_to_stream(plist_t plist, FILE* stream)
 		break;
 	default:
 		plist_node_print_to_stream(plist, &indent, stream);
+	}
+}
+
+int is_udid_valid(char *udid)
+{
+	if (!udid) {
+		return 0;
+	}
+
+	regex_t regex;
+	regcomp(&regex, "^[[:xdigit:]-]{25,40}$", REG_EXTENDED);
+	int result = regexec(&regex, udid, 0, NULL, 0);
+	regfree(&regex);
+
+	if (result == 0) {
+		return 1;
+	}
+	else {
+		return 0;
 	}
 }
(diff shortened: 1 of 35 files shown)


munimkazia:
It looks like device IDs are 24 digit, not 25. Atleast as per idevice_id -l output. It doesnt show the hyphen either.

O-kasso:
Curious, I have both an XS and an XS Max and their UDIDs (as reported by `idevice_id -l` on latest master) are `00008020-001024C81428002E` and `00008020-001D54560A88002E`

munimkazia:
Hmm, I may not have the latest master of libimobiledevice and all its dependencies. I get the same UDID format but without the hyphen. I'll try updating everything and report back.

munimkazia:
@O-kasso Just installed libplist, usbmuxd and libmuxd from source, and then installed libimobiledevice from your PR (and then killed usbmuxd process too). 
I am still getting device IDs without the hypen for XS and XS max. 

I am wondering if version of xcode or OS has something to do with this, but it sounds unlikely.

munimkazia:
I actually have my own PR which assumes 24 digits which I developed 2 days ago, but I didn't open the PR to the main project yet, because i am getting lockdown -2 errors on ideviceinfo, though idevicepair works. So I was wondering if I was doing something wrong, and if its related to this 24/25 digit issue.  https://github.com/munimkazia/libimobiledevice/pull/1

O-kasso:
@munimkazia The regex used to validate the UDID format has hyphens as optional, so both your scenario and mine should work fine. I don't think a ton of specificity is required for validating UDIDs, just as long as we're ensuring that whatever's getting passed in is sane.

O-kasso:
@munimkazia My bad, I misread and thought it was 25 chars or 26 with hyphen. Adjusted the regex range.

O-kasso:
@munimkazia good catch, added to `idevicepair`! Do you know how to get the attention of whoever can approve and merge PRs here?

luismarquesfreire:
Hi @O-kasso, maybe mentioning the most active guys - @nikias @hadess Can you merge this request? Thanks

hadess:
> Also, every issue or PR and every comment on any issue will result in an email notification. No need to mention me :)

And I have no commit rights, so no need to mention me either :)

Issue #715, opened 2018-10-06
idevicedebug debug_info not printing because debug_level not set


tlancina:
`idevicedebug` uses [`debug_info`](https://github.com/libimobiledevice/libimobiledevice/blob/b34e3435c21d06b3e5a4e7b5246fb6ddb6641a9f/common/debug.c#L81), but it seems that [`idevice_set_debug_level`](https://github.com/libimobiledevice/libimobiledevice/blob/b34e3435c21d06b3e5a4e7b5246fb6ddb6641a9f/tools/idevicedebug.c#L232) sets the `debug_level` of `libimobiledevice.dylib` (from `libinternalcommon.a`->`common/debug.c` ), and not the `debug_level` of the `idevicedebug` executable (from `common/debug.c`).

You can see the different memory locations for `debug_level` with `nm idevicedebug | grep debug_level` and `nm libimobiledevice | grep debug_level` respectively.

I'm not super familiar with C, but digging in a little, it seems that the encapsulation of the library code with `-fvisibility=hidden` is intentional.  So one workaround could be to also call `internal_set_debug_level` from `idevicedebug` so it sets its own `debug_level` (this works for me on Mac OS), since exposing `debug_info` through the API doesn't seem like a great idea.  That or removing `debug_info` in favor of something else.

PR #716, opened 2018-10-06
set `debug_level` for idevicedebug as well

tlancina:
Possible solution for https://github.com/libimobiledevice/libimobiledevice/issues/715.


jpstotz:
This patch makes the debug_info logging entries finally appear on the console. Thanks for the fix. 

I simply don't understand why such a simple and effective fix is molding here for more than a year :(

nikias:
The reason for this patch to sit here is that idevicedebug is mostly broken (which is currently being worked on by guys like you, thanks) so it doesn't get real attention from my side. However the proper fix will be to not use debug_info in idevicedebug but a different local symbol. I will handle that.

nikias:
See commit ce0868cd7c6c037b70bf5329fd5e8d0d5100576f.

PR #723, opened 2018-10-13
add erase function

zytzjx:
Erase All Content and Settings

### DIFF
--- a/tools/idevicebackup2.c
+++ b/tools/idevicebackup2.c
@@ -77,7 +77,8 @@ enum cmd_mode {
 	CMD_UNBACK,
 	CMD_CHANGEPW,
 	CMD_LEAVE,
-	CMD_CLOUD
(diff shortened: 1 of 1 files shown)


Blefish:
Make sure result_code is set, also check the error of mobilebackup2_send_message, it can fail if device is not activated.

Blefish:
Yes, I missed that.

I meant that when you currently send this command, the process returns 255, I guess it should be 0.

jginsburgn:
Why has this PR not been merged?

Blefish:
Maybe this is not the right place for it. It's the easiest to add the code for it here, but idevicebackup2 does not represent wiping in any meaningful way :).

nikias:
Yes. It doesn't belong there. Too many users will lose data because they expect this is a backup tool. There's no need for a mainstream tool like this
IMO. Closing it for now.

PR #728, opened 2018-10-26
Support WiFi devices in idevice_new

qmfrederik:
As a side-effect of the changes in libusbmuxd which bring better support for USB and WiFi connections, `usbmuxd_get_device_by_udid` only returns USB-connected devices.

Because `idevice_new` called into `usbmuxd_get_device_by_udid`, you would now get a scenario where `idevice_list` returns USB and WiFi-connected devices, but `idevice_new` would return E_NO_DEVICE when connecting to a device which is connect over WiFi only.

This PR changes the call from `usbmuxd_get_device_by_udid` to `usbmuxd_get_device` with ` DEVICE_LOOKUP_USBMUX | DEVICE_LOOKUP_NETWORK` so that USB connections are preferred, but a fallback to WiFi connections is possible.

### DIFF
--- a/src/idevice.c
+++ b/src/idevice.c
@@ -250,7 +250,7 @@ LIBIMOBILEDEVICE_API void idevice_set_debug_level(int level)
 LIBIMOBILEDEVICE_API idevice_error_t idevice_new(idevice_t * device, const char *udid)
 {
 	usbmuxd_device_info_t muxdev;
-	int res = usbmuxd_get_device_by_udid(udid, &muxdev);
+	int res = usbmuxd_get_device(udid, &muxdev, DEVICE_LOOKUP_USBMUX | DEVICE_LOOKUP_NETWORK);
 	if (res > 0) {
 		idevice_t dev = (idevice_t) malloc(sizeof(struct idevice_private));
 		dev->udid = strdup(muxdev.udid);


zytzjx:
usbmuxd_get_device(udid, &muxdev, DEVICE_LOOKUP_USBMUX | DEVICE_LOOKUP_NETWORK);
but I tested on Debian(Linux debian 4.9.0-8-amd64 #1 SMP Debian 4.9.130-2 (2018-10-27) x86_64 GNU/Linux)
ideviceinfo --domain com.apple.mobile.wireless_lockdown
BonjourFullServiceName: 18:f1:d8:df:17:9f@fe80::1af1:d8ff:fedf:179f._apple-mobdev2._tcp.local.
EnableWifiConnections: true
SupportsWifi: true
SupportsWifiSyncing: true
pc and iPhone connect the same router.

fosple:
When will this PR be merged @nikias, or are there still issues? Thanks a lot!

donie-inhance:
I notice this call uses the UDID of the device. How do we get this identifier? Can we query the wifi for devices? Can we have some info on how this works?

qmfrederik:
It only works for devices which have already paired with usbmuxd. You can enumerate those devices via idevice_id, or the equivalent API call.

sionicion:
So I built libimobiledevice with this PR, and I'm a little confused. Does more work need to be done before something like Flutter can take advantage? I'm able to see my iPad but it seems it can't forward the ports it needs to forward, it can only forward ports when the iPad is physically connected, not via network.

zytzjx:
> It only works for devices which have already paired with usbmuxd. You can enumerate those devices via idevice_id, or the equivalent API call.

it works fine. I tested via WIFI connect.

2rgman:
> > It only works for devices which have already paired with usbmuxd. You can enumerate those devices via idevice_id, or the equivalent API call.
> 
> it works fine. I tested via WIFI connect.

Hi @zytzjx , does it work for you in linux with open source usbmuxd or with Mac/Windows with Apple usbmuxd?
i'm trying to understand if wifi sync can work on linux and how does it detect network devices

Thanks!

qmfrederik:
This was implemented (although a bit differently) via 5390060ec90bed0f565698a389cebc57c09c3ac8.

PR #731, opened 2018-11-06
idevice: Use OpenSSL 1.0.2's thread API

neheb:
Should be more stable.


neheb:
Note that this kills support for OpenSSL 0.9.8. 1.0 and above is fine.

nikias:
> Note that this kills support for OpenSSL 0.9.8. 1.0 and above is fine.

This is exactly why I can't accept this pull request.

I committed a proper change that keeps it working for older OpenSSL versions.

neheb:
That works too.

PR #745, opened 2019-01-07
Update m4 macros

corsac-s:
Hi

here are few updates to the m4 folder, mainly to fix an issue with CPPFLAGS beeing erased and not restored.


nikias:
I merged the missing changes manually since I already did part of the patches in this PR before.

PR #766, opened 2019-03-28
libimobiledevice: Fix compilation without deprecated OpenSSL 1.1 APIs

neheb:
There are several missing headers as well as deprecated functions for
which compatibility was added.

### DIFF
--- a/common/userpref.c
+++ b/common/userpref.c
@@ -37,6 +37,7 @@
 #include <unistd.h>
 #include <usbmuxd.h>
 #ifdef HAVE_OPENSSL
+#include <openssl/bn.h>
 #include <openssl/pem.h>
 #include <openssl/rsa.h>
 #include <openssl/x509.h>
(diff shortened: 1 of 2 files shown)


neheb:
Rebased and added support for LibreSSL.

nikias:
Ok I merged this after a few small corrections:

`X509_set1_notBefore` and `X509_set1_notAfter` were introduced in OpenSSL 1.1.0 but were not present in OpenSSL 1.1.0-pre7-dev so I had to change the OPENSSL_VERSION to < 0x1010000fL.

`TLS_method` was introduced in LibreSSL 2.2.2, so I changed the version check to use it starting that version.

`SSL_COMP_free_compression_methods` was not introduced in LibreSSL 2.7.0, and is not even in libreSSL 2.9.2 yet, so I had to remove that version constraint.

neheb:
Great. Thanks for looking into it.

PR #767, opened 2019-03-28
configure.ac: Remove largefile hackery

neheb:
It requires a program to be ran, which cannot be done when cross compiling.
AC_SYS_LARGEFILE seems to be enough to implement largefile checking.


nikias:
You are right, AC_SYS_LARGEFILE seems totally sufficient.
Merged after rebase.

PR #792, opened 2019-06-01
common: Update thread.c to match one with libusbmuxd

s0uthwest:


nikias:
I already had the changes of the PR in my backlog of patches. Pushed it to master now so this PR is not needed anymore.

s0uthwest:
Ok, but if I’ll compile static versions of idevicerestore/futurerestore, I have errors with old thread.

PR #798, opened 2019-06-12
Timeout support for SSL connections and better timeout handeling.

DanyL:


nikias:
Merged after rebase. Thanks!

Issue #50, opened 2014-01-15
SIGPIPE on SSL_shutdown causes idevice commands to have non-zero exit code in Linux


greg-dennis:
To see a sympton of this, in bash try
  date=$(idevicedate) && echo $date
And see that at least 9 times out of 10, it produces no output. Echo $? to see the exit code is 141, not zero, indicating a SIGPIPE.

I have tracked it down the the SSL_shutdown call in idevice.c. I have worked around the issue by temporarily ignoring SIGPIPE signals when calling SSL_shutdown. Specifically, I've replaced the SSL_shutdown call on line 824 with:

  signal(SIGPIPE, SIG_IGN);
  SSL_shutdown(connection->ssl_data->session);
  signal(SIGPIPE, SIG_DFL);

It is unclear to me whether there is a bug here or whether ignoring the SIGPIPE is just standard when using openssl.

PR #847, opened 2019-09-18
ideviceimagemounter: fix crash in connection teardown

kamstrup:
We observed that ideviceimagemounter was consistently exiting with code 141 (tested ios9 and 13GM) and furthermore had no output when stdout was piped to a file.

Chased it down to the bug #50 issue. I tried all sorts of defensive measures in idevice_connection_disable_ssl(), but nothing worked. Ignoring SIGPIPE seems a bit like a band aid, but I can see that all other libimobile tools have it, so I guess it's ok :-)

### DIFF
--- a/tools/ideviceimagemounter.c
+++ b/tools/ideviceimagemounter.c
@@ -154,8 +154,10 @@ int main(int argc, char **argv)
 	size_t image_size = 0;
 	char *image_sig_path = NULL;
 
+#ifndef WIN32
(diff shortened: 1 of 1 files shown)


nikias:
Looks good, but please add an #ifndef WIN32  around it :)

kamstrup:
Done @nikias :+1:

nikias:
Solved with a larger commit that does it for all remaining tools.

Issue #793, opened 2019-06-04
idevicedebug iOS 13


maznikoff:
Unfortunately, something was broken with iOS 13 release. Tried to perform idevicedebug run bundle_id and got:

Jun  4 12:34:15 iPhone debugserver[417] <Notice>: debugserver will use ASL for internal logging.
Jun  4 12:34:15 iPhone debugserver[417] <Notice>: debugserver-@(#)PROGRAM:LLDB  PROJECT:lldb-900.3.96
 for arm64.
Jun  4 12:34:15 iPhone debugserver[417] <Notice>: Connecting to com.apple.debugserver service...
Jun  4 12:34:15 iPhone trustd[153] <Notice>: cert[0]: AnchorTrusted =(leaf)[force]> 0
Jun  4 12:34:15 iPhone trustd[153] <Notice>: cert[0]: NonEmptySubject =(path)[]> 0
Jun  4 12:34:15 iPhone debugserver[417] <Notice>: Got a connection, waiting for process information for launching or attaching.
Jun  4 12:34:15 iPhone lockdownd[67] <Notice>: spawn_xpc_service_block_invoke: <private>
Jun  4 12:34:15 iPhone lockdownd[67] <Notice>: spawn_xpc_service_block_invoke: <private>
Jun  4 12:34:15 iPhone debugserver[417] <Notice>: error: packet read thread exited.
Jun  4 12:34:15 iPhone debugserver[417] <Notice>: Exiting.

ZhaoWenHao1:
遇到了同样的问题

EliyahuStern:
@nikias, In iOS 13, most of the services require an extra SSL handshake once the service socket is connected (not just for the startService in lockdown client). I think this is the cause here too.

Dan-Maor:
What do you mean by "extra SSL handshake"?
Can you please elaborate?

EliyahuStern:
Lockdown return an extra key in response to StartService - EnableServiceSSL. When it is true, an SSL handshake is required on the new service socket after usbmuxd Connect.

nikias:
@EliyahuStern but this is actually implemented here: 
https://github.com/libimobiledevice/libimobiledevice/blob/master/src/lockdown.c#L1352
and here:
https://github.com/libimobiledevice/libimobiledevice/blob/master/src/service.c#L71
Is there anything else required?

EliyahuStern:
@nikias you are right, missed that somehow. I will take another look...

[...]

PR #850, opened 2019-09-27
idevice: fix hang in SSL_shutdown

kamstrup:
We observed that idevice_connection_disable_ssl() (ie. any call to idevice_disconnect) would frequently hang on ios13 devices when doing SSL bypass for the services that need it.

Relates to the discussion on #793


nikias:
Merged after rebase. Thanks!

Issue #118, opened 2014-09-15
iOS 7.1.2 Trust Loop Issue


oneguy:
Hi,

I'm having trust loop issue. I've compiled the libimobiledevice, but when i try to mount developers image here's what i get:
User@DEVPC /usr/local/bin
$ ideviceimagemounter.exe DeveloperDiskImage.dmg
ERROR: could not connect to lockdown. Exiting.

[here i see Trust dialog on the device. i click Trust. then try again:]

User@DEVPC /usr/local/bin
$ ideviceimagemounter.exe DeveloperDiskImage.dmg
ERROR: could not connect to lockdown. Exiting.

but the device reconnects and this process goes in a loop if i try again.

How do I troubleshoot this or if a fix is known, please let me know. Appreciate your help a lot!

PR #858, opened 2019-10-03
idevicebackup[2]: Replace usleep with nanosleep

neheb:
usleep was removed in POSIX 2008 and is optionally unavailable with
uClibc-ng.


mexmer:
sse my comment at #118

PR #859, opened 2019-10-10
debugserver: proxies SERVICE_E_TIMEOUT into DEBUGSERVER_E_TIMEOUT

dkimitsa:
this error is missing but required to properly handle timeout case in debugserver_client_receive_with_timeout


nikias:
Merged after rebase.

PR #860, opened 2019-10-10
ios13 debug server fix (#793)

dkimitsa:
this PR extends `idevice_connection_disable_ssl` with option to not send SSL shutdown but only free SSL related objects. 
it is required in case when SSL handshake is required by service but plain text communication is expected after it.  it is wrong to use `idevice_connection_disable_ssl` in this case as it will send SSL traffic over underlying socket and plain text service will receive SSL data and consider it as protocol data. E.g. as GDB one.

### DIFF
--- a/include/libimobiledevice/libimobiledevice.h
+++ b/include/libimobiledevice/libimobiledevice.h
@@ -241,6 +241,20 @@ idevice_error_t idevice_connection_enable_ssl(idevice_connection_t connection);
 idevice_error_t idevice_connection_disable_ssl(idevice_connection_t connection);
 
 /**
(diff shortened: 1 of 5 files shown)


michelleangela:
What is the status of this fix?

jpstotz:
From my perspective this fix does what it is supposed to to and I did not recognize any negative effects. 

@nikias: Is there something that contradicts applying this patch?

dcnt:
Hi,

Are there any chances to get this fix merged soon?

Thanks.

jpstotz:
Some days ago I had a conversation with @FunkyM and he promised to process some of the PRs including this in March.

nikias:
Merged after rebase. Thanks.

PR #896, opened 2019-12-29
tools/idevicebackup2: add --password-from-env option

tdyas:
Add `--password-from-env` option to idevicebackup2 so that the backup password can be obtained from an environment variable and not from a command-line parameter.

This should be slightly more secure in that the password won't end up in the shell's command history.

### DIFF
--- a/tools/idevicebackup2.c
+++ b/tools/idevicebackup2.c
@@ -1415,6 +1415,7 @@ static void print_usage(int argc, char **argv)
 	printf("    --remove\t\tremove items which are not being restored\n");
 	printf("    --skip-apps\t\tdo not trigger re-installation of apps after restore\n");
 	printf("    --password PWD\tsupply the password of the source backup\n");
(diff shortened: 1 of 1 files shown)


LaikaN57:
I do like being able to pass secrets over env vars though.

tdyas:
> I think the better pattern here is going to be, before the CLI args are parsed, to just look for particular env vars such as IDEVICEBACKUP2_PASSWORD (for example). 

I'm fine with that. I can switch the PR to use that method. Should the environment variables be `COMMAND_PASSWORD` so `IDEVICEBACKUP2_PASSWORD` and so forth?

LaikaN57:
Do we already use env vars in any other tools? If not, and we are setting pattern here, then I would:
1) Ask @nikias if they have any guidance/preference on naming here.
2) Do a quick survey of some popular software (eg. Python, GCC, Autoconf/Automake, Nginx, etc.) to see what their conventions are.

nikias:
See 1f625b2.

PR #906, opened 2020-01-24
idevicedebug lacks proper error handling

jpstotz:
The code of idevicedebug completely ignores that the device may fail to send a response within the configured time out.

idevicedebug ignores this situation which often works well but in other situations it leads to the problem that the app can't be started on the device but idevicedebug continues as if the app would run.
Therefore idevicedebug continues to run (unless the user manually kills the idevicedebug process).

For making idevicedebug working reliable this PR add proper error handling for all responses in the set-up phase.

Note: I strongly recommend also to increase the default time-out of 1000 milliseconds as it is a bit low from my experience. In my opinion idevicedebug should provide a parameter to set the time-out instead of having a built-in unchangeable value.

### DIFF
--- a/tools/idevicedebug.c
+++ b/tools/idevicedebug.c
@@ -352,6 +352,8 @@ int main(int argc, char *argv[])
 					}
 					free(response);
 					response = NULL;
+				} else {
+					goto cleanup;
 				}
 			}
 
@@ -370,6 +372,8 @@ int main(int argc, char *argv[])
 				}
 				free(response);
 				response = NULL;
+			} else {
+				goto cleanup;
 			}
 
 			/* set working directory */
@@ -386,6 +390,8 @@ int main(int argc, char *argv[])
 				}
 				free(response);
 				response = NULL;
+			} else {
+				goto cleanup;
 			}
 
 			/* set environment */
@@ -427,6 +433,8 @@ int main(int argc, char *argv[])
 				}
 				free(response);
 				response = NULL;
+			} else {
+				goto cleanup;
 			}
 
 			/* set thread */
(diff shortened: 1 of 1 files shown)


dkimitsa:
@jpstotz instead of checking response for null 
dres shall be checked against of DEBUGSERVER_E_SUCCESS as its explicit error code from API. 
there is also #859 that adds timeout error code to debug server, we use it in MobiVM project for proper error handling.

jpstotz:
@dkimitsa Thanks for pointing point that there is already a fix for that problem. I am not familiar with the debug server protocol, therefore I was not aware that your patch has the same effect as mine. 

But again there is the problem that an important PR that incorporate fixes remains unnoticed by the maintainer(s). At the moment the state of this project is a bit frustrating - it feels like a puzzle with a lot's of hidden parts (the PRs you have to apply for making the project actual work).

jpstotz:
In favor of #859 I close this pull request.

dkimitsa:
@jpstotz #859 doesn't solve issues you covered here. 
your PR is valid in respect to issue your mentioned. 
I just mentioned that instead of checking for response (which is indirect indicator of failure) 
```
if (response) {
} else { 
   goto cleanup;
}
```
its better to check for `dres` (which is direct status) like:
```
if (dres != DEBUGSERVER_E_SUCCESS) {
   // error, cleanup whatever
}
```

#859 just adds additional error code for timeout case.

jpstotz:
OK, I will check that. Additionally there is this new PR #914 which, if I understand it correctly, should also affect the error and time-out handling. 

But I have not tried neither your PR not #914 out. That is still on my todo list.

PR #914, opened 2020-02-11
Process LLDB stop reply packets correctly in idevicedebug

dogben:
This pull request makes a series of changes to fix the handling of LLDB reply packets in idevicedebug.

1. It is not allowed to reply to reply packets. Normally the server just ignores the extra commands, but sending the extra commands after a `W` or `X` reply causes problems because the connection has already been closed. Remove all `debugserver_client_send_command` from `debugserver_client_handle_response`.

2. It seems the only place that idevicedebug received the reply packets was in the call to `debugserver_client_send_command` in `debugserver_client_handle_response`. Instead, make the receive more explicit in the main loop.

3. In order to handle periods where there are no reply packets within the hard-coded time limit of 1 second, add support for looping until a response is received or a "`cancel_receive`" callback indicates we should stop waiting. (It might be better to implement this as a select rather than a poll, but that would involve plumbing the signal handler several layers down, which is a much bigger change than I would like to make.)

4. Fix handling of `W` reply and add support for `X` reply. If we see these codes, use the exit status of the app as the exit code of idevicedebug.

This pull request depends on and includes https://github.com/libimobiledevice/libimobiledevice/pull/859.


src/debugserver.c
```diff
@@ -83,6 +85,8 @@ LIBIMOBILEDEVICE_API debugserver_error_t debugserver_client_new(idevice_t device
 	debugserver_client_t client_loc = (debugserver_client_t) malloc(sizeof(struct debugserver_client_private));
 	client_loc->parent = parent;
 	client_loc->noack_mode = 0;
+	client_loc->cancel_receive = NULL;
+	client_loc->receive_loop_timeout = 1000;
```
dogben: These defaults should make the default behavior match the behavior before this change.

src/debugserver.c
```diff
@@ -152,7 +156,12 @@ LIBIMOBILEDEVICE_API debugserver_error_t debugserver_client_receive_with_timeout
 
 LIBIMOBILEDEVICE_API debugserver_error_t debugserver_client_receive(debugserver_client_t client, char* data, uint32_t size, uint32_t *received)
 {
-	return debugserver_client_receive_with_timeout(client, data, size, received, 1000);
+	debugserver_error_t res = DEBUGSERVER_E_UNKNOWN_ERROR;
+	do {
+		/* Is this allowed to return DEBUGSERVER_E_TIMEOUT and also set data and received? */
```
dogben: I don't see anywhere else in the code where this possibility is handled, so I didn't handle it here.

tools/idevicedebug.c
```diff
@@ -102,78 +107,63 @@ static instproxy_error_t instproxy_client_get_object_by_key_from_info_directiona
...
-		free(*response);
-		*response = NULL;
-
-		/* no command */
-		debugserver_command_new("OK", 0, NULL, &command);
-		dres = debugserver_client_send_command(client, command, response, NULL);
-		debug_info("result: %d", dres);
-		debugserver_command_free(command);
-		command = NULL;
+		debug_info("empty response");
```
dogben: I suspect this case was added to handle responses where the receive timed out. It may not be needed anymore.

dogben:
I've now tested this with iPhone 11 on iOS 13.2.3 and with iPhone 7 on iOS 11.4.1 using a C++ binary (packaged as an app) and it seems to work.

On the iPhone 11, both with and without this patch, I'm not able to interrupt idevicedebug (which kills the app) and then start the same app again. It seems to work fine on the iPhone 7.

dogben:
The last two commits fix a problem I observed where in some cases the connection to the debugserver is interrupted. In this case, I see the following in the logs:
14:15:37 idevice.c:552 internal_connection_receive_timeout(): ERROR: usbmuxd_recv_timeout returned -104 (Connection reset by peer)
8m 35s 14:15:37 service.c:167 service_receive_with_timeout(): could not read data
8m 35s 14:15:37 debugserver.c:148 debugserver_client_receive_with_timeout(): Could not read data, error -256

However, the error is actually ignored in debugserver_client_receive_internal_check and not propagated through debugserver_client_receive_response. Fixing this required quite a few changes to debugserver_client_receive_internal_check; I also found the logic in that function a bit confusing, so I simplified it.

We're now using this patch for all tests we run in our iOS test lab and I tested several failure cases, including non-zero exit status, C++ assertion failure, and segfault. Everything seems to be working as expected.

erikrose:
FWIW, as of the current tip of tree (fa8bfb65c70edd4d2617fbbf970302beb9a4ced2), our several dozen phones still need this patch in order for idevicedebug to communicate reliably.

nikias:
Merged after rebase @dogben (and also @erikrose) I would appreciate if you could get it a spin and see if it still works as expected with latest code from master.

PR #964, opened 2020-05-26
Fix parameter names in lockdown.h and afc.h

Schlaubischlump:
The parameter names in the documentation did not match the function parameter names.


nikias:
Correctly fixed with 6cd994d.

PR #974, opened 2020-06-12
Fix TLS version handling on Ubuntu 20.04

qmfrederik:
It appears that OpenSSL 1.1 in Ubuntu 20.04 is configured to reject older versions of TLS.

This causes `idevice_connection_enable_ssl` to fail with an `SSL_ERROR_SSL` error and the pairing record for that device to be deleted.

I've modified `ssl_error_to_string` to get the underlying SSL error message and you get this: `error:141E70BF:SSL routines:tls_construct_client_hello:no protocols available`

This PR modifies `idevice_connection_enable_ssl` to `SSL_CTX_set_security_level` and set the security level to 0, bringing back compatibility with OpenSSL 1.0.

### DIFF
--- a/src/idevice.c
+++ b/src/idevice.c
@@ -922,7 +922,7 @@ static const char *ssl_error_to_string(int e)
 		case SSL_ERROR_NONE:
 			return "SSL_ERROR_NONE";
 		case SSL_ERROR_SSL:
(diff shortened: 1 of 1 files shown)


nikias:
Merged with slight edit. Thanks!

PR #984, opened 2020-06-27
Fixed: Mistreatment of 0-byte sent cases

kakyoism:
Currently if 0 byte gets sent, it is treated as not-enough-data. 
This is wrong, because with TCP, 0-byte-sent usually means the receiver end is closed. We must set a new case for this and must not normalize the sent-bytes to 0 in general.


nikias:
Merged with 4a77478.

PR #1012, opened 2020-08-23
Resolves #1004: fallback to secured services

### LINKED ISSUE #1004: idevicedebug could not start to com.apple.debugserver on iOS 14.0 beta4 devices under Xcode 12.0 beta4
How to reproduce it
1. Use `xcode-select -s` to switch to Xcode 12.0 beta4.
2. Run `idevicediagnostics restart` to restart the device.
3. Run `idevicedebug run` to launch the application.
4. Failed with `Could not start com.apple.debugserver!`.
5. Switching to Xcode 12.0 beta3 and repeating step 2, 3 could succeed.

Btw, I could still use Xcode IDE to launch application and use `xcodebuild` to launch test on the same device under Xcode 12.0 beta4.

EliyahuStern:


src/debugserver.c
```diff
@@ -80,7 +80,12 @@ LIBIMOBILEDEVICE_API debugserver_error_t debugserver_client_new(idevice_t device
 		debug_info("Creating base service client failed. Error: %i", ret);
 		return ret;
 	}
-	service_disable_bypass_ssl(parent, 1);
+    
+    /* Disable service SSL only in iOS 13, where "com.apple.debugserver"
+     is not yet replaced by "com.apple.debugserver.DVTSecureSocketProxy" */
+    if (strcmp(service->identifier, "com.apple.debugserver")) {
```
dkimitsa, line 86: DEBUGSERVER_SERVICE_NAME instead of plain string ?

src/debugserver.c
```diff
@@ -80,7 +80,12 @@ LIBIMOBILEDEVICE_API debugserver_error_t debugserver_client_new(idevice_t device
 		debug_info("Creating base service client failed. Error: %i", ret);
 		return ret;
 	}
-	service_disable_bypass_ssl(parent, 1);
+    
+    /* Disable service SSL only in iOS 13, where "com.apple.debugserver"
+     is not yet replaced by "com.apple.debugserver.DVTSecureSocketProxy" */
+    if (strcmp(service->identifier, DEBUGSERVER_SERVICE_NAME)) {
```
dkimitsa, line 86: here is a bug: should be 
> if (strcmp(service->identifier, DEBUGSERVER_SERVICE_NAME) == 0)

src/lockdown.c
```diff
@@ -1274,39 +1305,39 @@ static lockdownd_error_t lockdownd_do_start_service(lockdownd_client_t client, c
...
+            send_start_service_receive_response(client, &dict, identifier, &ret, send_escrow_bag);
+        } else if (strcmp("com.apple.testmanagerd.lockdown", identifier) == 0) {
+            identifier = "com.apple.testmanagerd.lockdown.secure";
+            send_start_service_receive_response(client, &dict, identifier, &ret, send_escrow_bag);
+        } else if (strcmp(DEBUGSERVER_SERVICE_NAME, identifier) == 0) {
+            identifier = DEBUGSERVER_SECURED_SERVICE_NAME;
+            send_start_service_receive_response(client, &dict, identifier, &ret, send_escrow_bag);
+        }
+    }
+
```
dkimitsa, line 1334: if there is no `dict` at this point (e.g. LOCKDOWN_E_PLIST_ERROR) flow should not continue.
as it will cause GPF/NPE on accessing it and on plist_free

src/lockdown.c
```diff
@@ -1251,6 +1252,36 @@ static lockdownd_error_t lockdownd_build_start_service_request(lockdownd_client_
 	return LOCKDOWN_E_SUCCESS;
 }
 
+static void send_start_service_receive_response(lockdownd_client_t client, plist_t *dict, const char *identifier, lockdownd_error_t *ret, int send_escrow_bag) {
```
dkimitsa, line 1255: probably returning error instead of returning it by pointer would make code better readable

dkimitsa:
just as confirmation: after fixing a bug with `strcmp` it working for me.

nikias:
I have a similar change pending already. The handling code in `lockdown.c` is not how I would change it; rather the start service retry needs to go to `debugserver.c` respectively. I am doing it like this:

```diff
@@ -95,7 +98,11 @@ LIBIMOBILEDEVICE_API debugserver_error_t debugserver_client_new(idevice_t device
 LIBIMOBILEDEVICE_API debugserver_error_t debugserver_client_start_service(idevice_t device, debugserver_client_t * client, const char* label)
 {
        debugserver_error_t err = DEBUGSERVER_E_UNKNOWN_ERROR;
-       service_client_factory_start_service(device, DEBUGSERVER_SERVICE_NAME, (void**)client, label, SERVICE_CONSTRUCTOR(debugserver_client_new), &err);
+       service_client_factory_start_service(device, DEBUGSERVER_SECURE_SERVICE_NAME, (void**)client, label, SERVICE_CONSTRUCTOR(debugserver_client_new), &err);
+       if (err != DEBUGSERVER_E_SUCCESS) {
+               err = DEBUGSERVER_E_UNKNOWN_ERROR;
+               service_client_factory_start_service(device, DEBUGSERVER_SERVICE_NAME, (void**)client, label, SERVICE_CONSTRUCTOR(debugserver_client_new), &err);
+       }
        return err;
 }
```

Now since we also know the device product version we could also add a check for older versions and not try the secure one in the first place to save an extra request-reply cycle.

nikias:
Also kudos to @EliyahuStern for guessing some of my changes 🤣 
This is my current patch: https://gist.github.com/nikias/b2a669ef7cb44e448f7b6455317f64e6

dkimitsa:
once switched to SSL connection have discovered that SSL read doesn't honour timeouts correctly, opened the issue (with possible fix proposed) https://github.com/libimobiledevice/libimobiledevice/issues/1016

nikias:
I pushed commit 98056a89648f431759c5fa4ed87c6ea6ba0cdd3f which solves it for debugserver. For any other service it has to be handled separately like I did with debugserver.

EliyahuStern:
@nikias so it was you all along 😄 
> IDE Device Support
> Available for: iPhone 6s and later, iPod touch 7th generation, iPad
> Air 2 and later, and iPad mini 4 and later
> Impact: An attacker in a privileged network position may be able to
> execute arbitrary code on a paired device during a debug session over
> the network
> Description: This issue was addressed by encrypting communications
> over the network to devices running iOS 14, iPadOS 14, tvOS 14, and
> watchOS 7.
> CVE-2020-9992: Dany Lisiansky (@DanyL931), Nikias Bassen

You could have fixed this in advance...

nikias:
🙃

PR #1050, opened 2020-10-22
idevicedebug: allow to start an app and exit idevicedebug without killing the started app

jpstotz:
I often have the use case where I want to start an app on the iPhone and then disconnect the debugger from that app. 

As idevicedebug by default kills the started app when it ends (a feature that was not working correctly for a long time) I added a new command-line option `--disconnect` that starts an app disconnects the debugger and ends idevicedebug. In this case the app will continue to run.


nikias:
Looks good to me regarding the functionality, but maybe we can change the term `disconnect` to the more accurate `detach`.
```
 --detach		detach from app after launch, keeping it running
```

jpstotz:
@nikias Thanks for your feedback. I have changed the command, help text, debug text and used variable as you suggested.

nikias:
I squashed the commits and merged after rebase.

PR #1079, opened 2020-12-04
Fix typo found in README

youngchief-btw:
It said "serviers" instead of "servers"


nikias:
Fixed with 07485a5bf24ebce4d915510594704fcab42e39ec.

PR #1087, opened 2020-12-11
Fix idevicescreenshot build on Ubuntu

qmfrederik:
be72ca64cd9e853588ec65047d16a0a22f4ef291 added calls to `floor` and `log` which are defined in `libm`.

Check for this function and link `idevicescreenshot` with `libm`.

Fixes the build on Ubuntu.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -62,6 +62,9 @@ if test "x$ac_cv_have_endian_h" = "xno"; then
   fi
 fi
 
+# idevicescreenshot uses floor
+AC_CHECK_LIB([m],[floor])
+
 # Check for operating system
 AC_MSG_CHECKING([for platform-specific build settings])
 case ${host_os} in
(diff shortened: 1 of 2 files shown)


qmfrederik:
/cc @progressnerd , without this patch, master fails like this:

```
make[1]: Entering directory '/__w/1/s/libimobiledevice/tools'
  CCLD     idevicescreenshot
/usr/bin/ld: idevicescreenshot-idevicescreenshot.o: undefined reference to symbol 'floor@@GLIBC_2.2.5'
```

nikias:
Obsolete with commit 9c3dd915ea00ce33d6dee359bcef963a586aaf05 which removes `floor()` and `log10()` and also improves the memory usage since it doesn't free/malloc in every loop iteration anymore.

qmfrederik:
Thanks, @nikias

PR #1098, opened 2020-12-24
clang-tidy: use const pointer parameter

neheb:
Found with readability-non-const-parameter


nikias:
Merged with 3de1317.

PR #1100, opened 2020-12-24
clang-tidy: fix bad floating point cast

neheb:
Found with bugprone-integer-division


LaikaN57 (approved):

LaikaN57:
Good catch. Small, but good to still have correct code.

nikias:
Merged with 7804088.

PR #1103, opened 2020-12-24
clang-tidy: do not use else after return

neheb:

### DIFF
--- a/common/socket.c
+++ b/common/socket.c
@@ -398,14 +398,13 @@ static int32_t _sockaddr_in6_scope_id(struct sockaddr_in6* addr)
 			if (addr->sin6_scope_id == addr_in->sin6_scope_id) {
 				res = addr_in->sin6_scope_id;
 				break;
-			} else {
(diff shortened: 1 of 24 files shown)


nikias:
Merged with cf2c7d48380969200034afeab93169e68c34c397.

mexmer:
i checked patch to common/socket.c and there is problem, well problem is with function _sockaddr_in6_scope_id
when using getifaddrs you are not guaranteed addresses to be ordered by scope_id, original implementation was wrong, and "clang fixed" implementation just made this error more obvious.

while this
```
			if (addr->sin6_scope_id == addr_in->sin6_scope_id) {
				res = addr_in->sin6_scope_id;
				break;
			}
```
is and was correct.

this
```
			if ((addr_in->sin6_scope_id > addr->sin6_scope_id) && (res >= 0)) {
				// use last valid scope id as we're past the requested scope id
				break;
			}
```
was incorrect from start

therefore you should iterate trough all item, until you find one that matches

PR #1119, opened 2021-02-17
Remove redundent `select` call from idevicedebugserverproxy.c

nisargjhaveri:
- Turns out `socket_receive` anyway internally calls `select` again. Use `socket_receive_timeout` instead.
- Remove logging for timeout in socket.c as well, it is not really an error condition

### DIFF
--- a/common/socket.c
+++ b/common/socket.c
@@ -1084,8 +1084,6 @@ int socket_check_fd(int fd, fd_mode fdm, unsigned int timeout)
 				return -1;
 			}
 		} else if (sret == 0) {
-			if (verbose >= 2)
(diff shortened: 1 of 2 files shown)


nisargjhaveri:
@nikias as you've done changes here recently.

One more thing here, since we're not able to select anyway here and are waiting for the two fds separately, should we go back to multiple threads to save some cycles?

nikias:
Oh good catch. I meant to use read() instead of socket_receive_timeout, but looks like I forgot to change it. I tried to measure the throughput and it seemed like with the select method it runs faster. Removing the select is obviously on spot here, since we don't need it to call select twice.

PR #1124, opened 2021-03-04
Update afc.c

xdeng:
When the service_receive function fails, the size of bytes_recv will be fatal.


nikias:
Solved via 10003bcd970eed414b09eb567d36facbfe92cecc.

PR #1125, opened 2021-03-08
Add bluetooth packet logger 

doggkruse:
Add implementation for com.apple.bluetooth.BTPacketLogger service to capture Bluetooth HCI traffic.  Includes tool 'btpacketlogger' to capture to pcap format that can be read with tools such as Wireshark.  Requires Bluetooth logging profile to be installed on the target device. https://developer.apple.com/bug-reporting/profiles-and-logs/?name=bluetooth

### DIFF
--- a/src/bt_packet_logger.c
+++ b/src/bt_packet_logger.c
@@ -1,0 +1,235 @@
+/*
+ * bt_packet_logger.c
+ * com.apple.bt_packet_logger service implementation.
+ *
+ * Copyright (c) 2021 Geoffrey Kruse, All Rights Reserved.
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+
+#ifdef HAVE_CONFIG_H
+#include <config.h>
+#endif
+#include <string.h>
+#include <stdlib.h>
+
+#include "bt_packet_logger.h"
+#include "lockdown.h"
+#include "common/debug.h"
+
+struct bt_packet_logger_worker_thread {
+	bt_packet_logger_client_t client;
+	bt_packet_logger_receive_cb_t cbfunc;
+	void *user_data;
+};
+
+/**
+ * Convert a service_error_t value to a bt_packet_logger_error_t value.
+ * Used internally to get correct error codes.
+ *
+ * @param err An service_error_t error code
+ *
+ * @return A matching bt_packet_logger_error_t error code,
+ *     BT_PACKET_LOGGER_E_UNKNOWN_ERROR otherwise.
+ */
+static bt_packet_logger_error_t bt_packet_logger_error(service_error_t err)
+{
+	switch (err) {
+		case SERVICE_E_SUCCESS:
+			return BT_PACKET_LOGGER_E_SUCCESS;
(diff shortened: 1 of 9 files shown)


mringwal:
Thanks for providing this! 

A friend asked me how to reverse engineer some weird BLE device. By getting access to the HCI trace (also without a Mac), this becomes much easier compared to logging the GATT requests of the app (jailbreak only), or getting an air-trace (which cannot be broken if LE Secure Connections are used). 

@nikias could you have another look at this PR? I've tried the tool and got a valid Bluetooth .pcap file.

src/bt_packet_logger.c
```diff
@@ -0,0 +1,235 @@
...
+
+LIBIMOBILEDEVICE_API bt_packet_logger_error_t bt_packet_logger_start_capture(bt_packet_logger_client_t client, bt_packet_logger_receive_cb_t callback, void* user_data)
+{
+	if (!client || !callback)
+		return BT_PACKET_LOGGER_E_INVALID_ARG;
+
+	bt_packet_logger_error_t res = BT_PACKET_LOGGER_E_UNKNOWN_ERROR;
+
+	if (client->worker) {
+		debug_info("Another syslog capture thread appears to be running already.");
```
mringwal, line 197: I guess this should be 'Another Bluetooth capture thread appears to be running already"

mringwal:
@doggkruse Would it be ok for you to use Apple's PacketLogger format instead of pcap? 

The benefit would be that Apple's PacketLogger tool can be used  as well while it still can be opened with Wireshark.

 I can help / implement that if you like.  Example code can be found here: https://github.com/bluekitchen/btstack/blob/master/src/hci_dump.c#L125

src/bt_packet_logger.c
nikias: @doggkruse ?

nikias:
Looks pretty cool actually! I wanted to merge this back then but then forgot about it. @mringwal maybe it can be implemented in a away that it supports both formats?

mringwal:
> maybe it can be implemented in a away that it supports both formats?

Sure. If the simple "just use PacketLogger as it can be opened with Wireshark as well" is not sufficient, there could be a command line option to pick a format if needed.

nikias:
Oh I missed that line, guess I should read on a bigger screen 🤣

doggkruse:
> @doggkruse Would it be ok for you to use Apple's PacketLogger format instead of pcap?
> 
> The benefit would be that Apple's PacketLogger tool can be used as well while it still can be opened with Wireshark.
> 
> I can help / implement that if you like. Example code can be found here: https://github.com/bluekitchen/btstack/blob/master/src/hci_dump.c#L125

My original use case was to pipe live to Wireshark.  Can the packet logger format be used with pipes?  I like the idea of supporting either format

mringwal:
> My original use case was to pipe live to Wireshark. Can the packet logger format be used with pipes? I like the idea of supporting either format
I wouldn't know for sure, but it looks like piping to Wireshark only allows for the pcap format: https://wiki.wireshark.org/CaptureSetup/Pipes

To keep your original use case, we could have libpcap as default file format with an option for PacketLogger (or BTSnoop, or BlueZ hcidump, ....)

mringwal:
@doggkruse btw... looking at your code to convert the data received from the ios service into pcap format. The received data seems to be 100% Apple PacketLogger format - which isn't surprising. After all, it's the format Apple's macOS logger has used all the time.

If we just pipe that into the file, it's a valid .pklg file. There's no file header btw. just a list of HCI packets.

mringwal:
Hi. I've added a --fromat option here: https://github.com/mringwal/libimobiledevice/tree/add-bt-packet-logger-formats
@doggkruse what do you think about it?

mringwal:
More details: I've rebased my version on master (using libimobiledevice-glue includes), selected PacketLogger as the native output format and updated the man page accordingly. Ready to go from my side.

mringwal:
@nikias the current code to save in pcap format uses libpcap-dev - which isn't listed as requirement yet.

 As pcap files are rather simple, would you prefer to add libpcap-dev as build requirement or write pcap format manually (its a header, which could be hard coded)?

mringwal:
@nikias To get this done, I've removed the dependency on libpcap by dropping support for the BTsnoop/PCAP file format. It now only stores the data from iOS into a file in PacketLogger format, which can be analyzed with Wireshark, too.

GitHub says there are "conflicts that must be resolved", but I've just rebased it on master.

nikias:
Superseded by #1292.

PR #1158, opened 2021-05-11
Fixed DEBUGSERVER_SECURE_SERVICE_NAME macro

benjtq:
secure service name shady, which causes SSL connections (idevicedebug) to always fail. Fixed it with this.


mexmer:
this patch doesn't make sense.
if original code is not working, then problem lies within your C++ preprocessor that is not properly concatenating stuff.

what compiler you use on what platform?

nikias:
@benjiburgess I have to agree with @mexmer here, it must be something wrong with your C compiler. This is a perfectly legal usage of macro concatenation.

PR #1164, opened 2021-05-14
Fix linking errors on Windows by linking to IPHLPAPI.DLL

nisargjhaveri:


nikias:
This is obsolete now since libimobiledevice-glue links against `Iphlpapi.dll` on Windows.

PR #1169, opened 2021-05-19
Update mobilesync version to support iOS 14

darshilgada24:
Mobilesync does not work for iOS 14. It fails with error message as `MOBILESYNC_E_BAD_VERSION`.

Changing the MSYNC_VERSION_INT1 to 400 would solve this issue.

### DIFF
--- a/src/mobilesync.c
+++ b/src/mobilesync.c
@@ -34,7 +34,7 @@
 #include "device_link_service.h"
 #include "common/debug.h"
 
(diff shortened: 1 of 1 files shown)


darshilgada24:
@nikias Can you please have a look at this PR?

nikias:
Fixed with commit ca324155f8b33babf907704828c7903608db0aa2.

PR #1170, opened 2021-05-23
idevicecrashreport windows file names

zecops-team:
Some files cannot be extracted on windows since they contain `:` in their name (mostly seen as part of a date format such as `hh:mm:ss` as part of the file name).

This is a small patch to make sure we don't miss those files.


nikias:
Merged with commit d0136024c69167a98343710fa6fcdc5843851006.

PR #1180, opened 2021-06-27
Added milliseconds in ios log

ndatta41:


guocdfeifei:
I also want this one with millisecond function. How can I install it with brew?

nikias:
The milliseconds for the debug output is fine; but the date+time is reflecting whatever the device is sending and I cannot just change that. There's a newer service called `os_trace_relay` that I am looking into which has microseconds resolution though.

j0rdanit0:
Any progress on this? I could also use millisecond precision.

leelindsaykatts (approved):

tools/idevicesyslog.c
```diff
@@ -354,7 +356,14 @@ static void syslog_callback(char c, void *user_data)
 
 				/* write date and time */
 				TEXT_COLOR(COLOR_DARK_WHITE);
-				fwrite(line, 1, 16, stdout);
+				char* str_time = (char*)malloc(15);
+				struct timespec t;
+				timespec_get(&t, 1);
```
xingheng, line 361: This is hard, if you use the time data sent from iOS device log, it's acceptable for me.

nikias:
I added milliseconds to the debug messages with 796d09d32de51e933bb300d9847c46e3a96eb9ac. But I will not add this to idevicesyslog, as the timestamp comes from the device and it should stay that way. I will soon publish os_trace_relay changes, that already comes with microseconds for the timestamps.

nikias:
Check out d3aec6d

PR #1210, opened 2021-09-03
Added missing dependencies for idevicepair.

RodoMa92:
The build on my system was failing with:

/usr/bin/ld: ../src/.libs/libimobiledevice-1.0.so: undefined reference to symbol 'string_concat'
/usr/bin/ld: /usr/lib/gcc/x86_64-pc-linux-gnu/11.1.0/../../../../lib/libimobiledevice-glue-1.0.so: error adding symbols: DSO missing from command line
...
make[2]: *** [Makefile:762: idevicepair] Error 1

After adding the new glue code to the C and LD flags the project now build correctly.


nikias:
Fixed it with 46869ab79dbc8828fea9f9807be0938fc211d6e5.

PR #1235, opened 2021-10-27
Win32: add MSVC compatible call

xanadu-lee:
This PR is similar to some release methods added by libplist.
On windows, exceptions occur when memory is released across dynamic libraries, especially different compilers.

### DIFF
--- a/include/libimobiledevice/debugserver.h
+++ b/include/libimobiledevice/debugserver.h
@@ -234,6 +234,14 @@ debugserver_error_t debugserver_command_free(debugserver_command_t command);
  */
 void debugserver_encode_string(const char* buffer, char** encoded_buffer, uint32_t* encoded_length);
 
+
+/**
+ * Release the encode string created by debugserver_encode_string
+ *
+ * @param ncoded_buffer Encoding string of debugserver_encode_string request
+ */
+void debugserver_encode_string_free(char* encoded_buffer);
+
 /**
  * Decodes a hex encoded string.
  *
@@ -242,6 +250,14 @@ void debugserver_encode_string(const char* buffer, char** encoded_buffer, uint32
  * @param buffer Decoded string to be freed by the caller
  */
 void debugserver_decode_string(const char *encoded_buffer, size_t encoded_length, char** buffer);
+
+
+/**
+ * Release the decode string created by debugserver_decode_string
+ *
+ * @param buffer Decoding string of debugserver_decode_string request
+ */
+void debugserver_decode_string_free(char* buffer);
 
 #ifdef __cplusplus
 }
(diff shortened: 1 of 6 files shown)


xanadu-lee:
Similar to plist_to_xml_free().
When a program compiled by MSVC calls the library, the memory cannot be released.

mexmer:
such functions are unnecessary, you call free(buffer), inside your code after you not need value anymore, introducing function that only does "free" and nothing else is just waste of calls. if anything, then to both encode and decode function description should be added line "allocates **buffer**, caller is responsible to release it".

xanadu-lee:
> such functions are unnecessary, you call free(buffer), inside your code after you not need value anymore, introducing function that only does "free" and nothing else is just waste of calls. if anything, then to both encode and decode function description should be added line "allocates **buffer**, caller is responsible to release it".

You know, not everyone is using Linux or MacOS.
If I use the dynamic library compiled by MinGW (I mainly use MSVC), an exception will occur.
I hope to follow the principle of who applies for release.

mexmer:
> > such functions are unnecessary, you call free(buffer), inside your code after you not need value anymore, introducing function that only does "free" and nothing else is just waste of calls. if anything, then to both encode and decode function description should be added line "allocates **buffer**, caller is responsible to release it".
> 
> You know, not everyone is using Linux or MacOS. If I use the dynamic library compiled by MinGW (I mainly use MSVC), an exception will occur. I hope to follow the principle of who applies for release.

i use libimobile on windows (compiled own msvc project), no exception occurs, when i call free, inside function, that had called encode or decode.

anyway, you function only decorates "free" call, does not do any safety check, or anything ... so if you get exception, problem lies elsewhere.

on plist, free function has meaning, because you release structure with children, and you need to release everything in cascade.

xanadu-lee:
> 

It should be noted that releasing an empty finger will not cause an exception. So there is no need for any checks. 
I use MinGW32 to compile libimobiledevice and other projects.
Then I use MSVC to compile my own project and call it. 
I will receive an exception when calling free(). 
But after I added the wrapper, everything worked fine. 
Just like libplist.
Why does the author need to add plist_to_xml_free()?

nikias:
It looks really bad having tons of *_free methods in the API interface. I initially didn't even want to add plist_to_xml_free() and plist_to_bin_free() for the same reason. In a recent commit I merged it into a single plist_mem_free() because otherwise I would have had to add like 3 or 4 additional _free functions to the interface.
The only thing I can offer is a single
libimobiledevice_mem_free().

xanadu-lee:
> It looks really bad having tons of *_free methods in the API interface. I initially didn't even want to add plist_to_xml_free() and plist_to_bin_free() for the same reason. In a recent commit I merged it into a single plist_mem_free() because otherwise I would have had to add like 3 or 4 additional _free functions to the interface. The only thing I can offer is a single libimobiledevice_mem_free().

That's enough. But I declare that it's not for me, because I can add it and compile it myself. I'm sorry to put you on the spot.

Issue #1231, opened 2021-10-21
idevice_event_subscribe overwrites existing callback


tihmstar:
When calling `idevice_event_subscribe` multiple times, the later invocations overwrite previous callbacks.
See: https://github.com/libimobiledevice/libimobiledevice/blob/master/src/idevice.c#L236

This becomes problematic, when in a bigger project for example multiple libraries rely on this functionality of libimobiledevice. In that case invocation of the second lib would straight up break the functionality of the first lib (expecially if libimobiledevice is linked dynamically).

In constrast libirecovery does handle this correctly by having a context (https://github.com/libimobiledevice/libirecovery/blob/master/src/libirecovery.c#L2367).

I suggest adding a similar concept to libimobiledevice such that multiple callbacks (eg. in different libraries) can be registered.

tihmstar:
I suspect this is resolved now?

venoj:
Is idevice_event_subscribe a standalone command that can be run directly from the terminal, and if so, what are the additional steps or dependencies required to enable it?

mexmer:
no it's api function

PR #1262, opened 2022-01-02
added context for idevice events

tihmstar:
Implementing contexts for callbacks as discussed in https://github.com/libimobiledevice/libimobiledevice/issues/1231


nikias:
Resolved with 2b48c417c5b4f2fa78ab0960eb718ff081fa48e6.

PR #1279, opened 2022-03-01
Fixed: Blocked in process when device restarted.

xanadu-lee:
When used with multiple threads, if a thread restarts the device.
Just when I was about to call the diagnostics_relay series of functions, diagnostics_relay_send had failed, but it did not check the return value, but read it directly, which would cause diagnostics_relay_receive to not return.


xanadu-lee:
This operation is not yet found on Mac and Linux, only Windows will occur.
The main reason is that when the device is restarted, the driver does not immediately notify the device to disconnect.
Actually the socket is invalidated, possibly blocking in SSL_read.

nikias:
I pushed this with commit 40b7451502fcad7041521346d701f77c9dc2deaf.

PR #1285, opened 2022-03-10
Use synchronous mode when callbacks are NULL

marcprux:
Synchronous modes for `instproxy_install`, `instproxy_upgrade`, `instproxy_uninstall`, `instproxy_archive`, `instproxy_restore`, and `instproxy_remove_archive` appear to be ignoring the documented behavior that “If NULL is passed, this function will run synchronously”.


nikias:
Merged with dd8a2de8e0bb8e2e81a83234db524a07969b6ad3.

PR #1292, opened 2022-03-26
Add bluetooth packet logger (2)

mringwal:
This is an extension of https://github.com/libimobiledevice/libimobiledevice/pull/1125

I've realized that while I've mentioned my code additions, I never provided a full link / pr for that, so here is it now. 
It comes in two flavours: with and without support for BTsnoop/PCAP files. The latter requires a dependency on libpcap, that's why the last commit rmoves support for libpcap to make it easier to merge this without much hassle.

If the last commit is dropped/ignored, BTsnoop/PCAP file format is supported but this requires a new depedency.

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,6 +128,7 @@ The library bundles the following command-line utilities in the tools directory:
 | `idevice_id`               | List attached devices or print device name of given device         |
 | `idevicebackup`            | Create or restore backup for devices (legacy)                      |
 | `idevicebackup2`           | Create or restore backups for devices running iOS 4 or later       |
+| `idevicebtlogger`          | Capture Bluetooth HCI traffic from a device (requires log profile) |
 | `idevicecrashreport`       | Retrieve crash reports from a device                               |
 | `idevicedate`              | Display the current date or set it on a device                     |
 | `idevicedebug`             | Interact with the debugserver service of a device                  |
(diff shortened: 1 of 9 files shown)


nikias:
@mringwal can Wireshark read PacketLogger format?

mringwal:
> @mringwal can Wireshark read PacketLogger format?

Yes. (Both file formats are eventually processed by the same HCI dissectors.)

mringwal:
yupp. live feedback only seems possible via PCPA/BTsnoop format :(

Choices:
- drop PCAP and live-feedback
- add libpcap dependency
- implement PCAP format without libpcap (some work)

nikias:
@mringwal deal :)
I would even be ok with a configure time selection of PCAP/no PCAP support depending on the availability of libpcap, but then there will always be tons of complaints about stuff not working, so I prefer your offer :)

mringwal:
Ok, stay tuned. I guess I'll have something towards the end of the week.

nikias:
No rush, take your time!

mringwal:
@nikias Probably ready now. I've tested logging into PCAP file for Wireshark and also managed to actually pipe the output into Wireshark for live feedback. See man.

nikias:
@mringwal great stuff! @doggkruse I'd like to put your full name and email into the commits; I DMed you on Twitter, so kindly respond :)

nikias:
Merged with df1da523f76fb03a9a73abdfc07a366331009819.

PR #1331, opened 2022-06-20
Windows replaces ssize_t with SSIZE_T

xanadu-lee:
When compiling with MSVC on Windows, there is no ssize_t type.
But there is a type with the same name capitalized, so we use conditional compilation for now.

### DIFF
--- a/include/libimobiledevice/mobile_image_mounter.h
+++ b/include/libimobiledevice/mobile_image_mounter.h
@@ -30,6 +30,9 @@ extern "C" {
 
 #include <libimobiledevice/libimobiledevice.h>
 #include <libimobiledevice/lockdown.h>
+#if defined(_WIN32)
+#include <windows.h>
+#endif
 
 /** Service identifier passed to lockdownd_start_service() to start the mobile image mounter service */
 #define MOBILE_IMAGE_MOUNTER_SERVICE_NAME "com.apple.mobile.mobile_image_mounter"
@@ -49,7 +52,11 @@ typedef struct mobile_image_mounter_client_private mobile_image_mounter_client_p
 typedef mobile_image_mounter_client_private *mobile_image_mounter_client_t; /**< The client handle. */
 
 /** callback for image upload */
+#if defined(_WIN32)
+typedef SSIZE_T (*mobile_image_mounter_upload_cb_t) (void* buffer, size_t length, void *user_data);
+#else
 typedef ssize_t (*mobile_image_mounter_upload_cb_t) (void* buffer, size_t length, void *user_data);
+#endif
 
 /* Interface */
 


mexmer:
SSIZE_T is MSVC specific, not windows platform specific. this will not compile under msys2 or cygwin.

mexmer:
how are you compiling it, because this should not be needed under cygwin/msys, in fact it will create problems when using cmake.

also SSIZE_T is not defined in windows.h, if you want to go with SSIZE_T, you should include <BaseTsd.h> instead, to minimize included header.

xanadu-lee:
I use Mingw32 to compile libimobiledevice.
Invoke this library using MSVC-CMake

xanadu-lee:
When I use MinGW to compile the library, using MSVC's program calls, the situation is different.
MinGW doesn't define MSC_VER, it defines WIN32.
It causes MinGW to use ssize_t, but MSVC to use SSIZE_T.
Detailed output is in the file below.

As for <Windows.h> vs <BaseTsd.h> , I think you're right.
Just out of habit, under Windows platform, <Windows.h> is generally included.

Are none of you using MSVC?
[mingw-gcc.txt](https://github.com/libimobiledevice/libimobiledevice/files/8964956/mingw-gcc.txt)

mexmer:
for MSVC compilation i created visual studio project, and link libimobiledevice,libplist,libimobiledevice-glue as static libs, but most time i compile inside msys2 with gcc, because that's easiest to get this working on windows.

xanadu-lee:
I think you don't quite understand what I mean.
You can't ask people to force the use of static libraries on MSVC.
For example, I can only use it on MinGW (Windows platform) at present, but my other projects are compiled by MSVC, including linking the QT library.

mexmer:
@xanadu-lee i didn't say, people need to use static libraries, i said, that for compilation using MSVC, i created visual studio project, and that one creates static libraries (of course it can create dll too, it was just easier and faster to setup static library vs project).

as for your situation, best option would be, to use config.h and put
```
#include <BaseTds.h>
typedef SSIZE_T ssize_t;
```
inside config.h
and compile with -DHAVE_CONFIG_H

xanadu-lee:
> @xanadu-lee i didn't say, people need to use static libraries, i said, that for compilation using MSVC, i created visual studio project, and that one creates static libraries (of course it can create dll too, it was just easier and faster to setup static library vs project).
> 
> as for your situation, best option would be, to use config.h and put
> 
> ```
> #include <BaseTds.h>
> typedef SSIZE_T ssize_t;
> ```
> 
> inside config.h and compile with -DHAVE_CONFIG_H

OK, no matter what the process is, as long as it can be used.

mexmer:
using config.h is more universal, although config.h is created by configure script, so if you don't use configure, you can use config.h.in to check, what you actually need to set (that's what i did with my visual studio project, i copied config.h.in to config.h, then edited entries i need, and added aditional entries specific for visual studio build, inside msys environment i use configure)

xanadu-lee:
If you can, you can fix it on your side.
I don't quite understand your rules.

PR #1352, opened 2022-09-04
Fix iOS build

kabiroberai:
This PR includes three orthogonal changes that fix building for iOS:

- Allow using limd as a submodule (relevant to use cases where limd is inside a parent project)
- Support OpenSSL built without OPENSSL_ENGINE (the [most popular](https://github.com/krzyzanowskim/OpenSSL) OpenSSL dist for iOS is built with `OPENSSL_NO_ENGINE`)
- Don't try to use `SCDynamicStoreCopyComputerName` on iOS – it's only defined in the macOS SDK.

### DIFF
--- a/git-version-gen
+++ b/git-version-gen
@@ -3,7 +3,7 @@ SRCDIR=`dirname $0`
 if test -n "$1"; then
   VER=$1
 else
-  if test -d "${SRCDIR}/.git" && test -x "`which git`" ; then
+  if test -r "${SRCDIR}/.git" && test -x "`which git`" ; then
     git update-index -q --refresh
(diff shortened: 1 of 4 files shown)


git-version-gen
```diff
@@ -3,7 +3,7 @@ SRCDIR=`dirname $0`
 if test -n "$1"; then
   VER=$1
 else
-  if test -d "${SRCDIR}/.git" && test -x "`which git`" ; then
+  if test -r "${SRCDIR}/.git" && test -x "`which git`" ; then
```
kabiroberai, line 6: this patch should be applied to libplist and libusbmuxd as well; @nikias I can put up PRs there if you want, or you could just push to master directly, whichever is more convenient for you.

PR #1393, opened 2022-12-20
Add timeout option for mobilebackup2_receive_message

AiXanadu:
Sometimes we may need to respond quickly, and the original 30-second waiting time is too long.
A new interface is added that allows the caller to customize the timeout period.

### DIFF
--- a/include/libimobiledevice/mobilebackup2.h
+++ b/include/libimobiledevice/mobilebackup2.h
@@ -128,6 +128,27 @@ mobilebackup2_error_t mobilebackup2_send_message(mobilebackup2_client_t client, 
 mobilebackup2_error_t mobilebackup2_receive_message(mobilebackup2_client_t client, plist_t *msg_plist, char **dlmessage);
 
 /**
+ * Receives a DL* message plist from the device with the specified timeout.
+ * This function is a wrapper around device_link_service_receive_message_with_timeout.
+ *
+ * @param client The connected MobileBackup client to use.
+ * @param msg_plist Pointer to a plist that will be set to the contents of the
+ *    message plist upon successful return.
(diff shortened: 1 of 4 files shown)


mexmer:
this interface is good addon, although i don't understand your reasoning.

if you want to cancel operation, you would want implement canceling callback, and not shortening timeout.

still i appreciate this function, atm i use modified version, that has 60 seconds timeout, because on some phones i testesd, initial reception (before or after manifest datbase is sent) takes up to 50 seconds, since it was only handfull of devices i never submitted PR for that.

mind timeout is not how long it waits in function, but how long it waits for response from device, if response from device is immediate, it will return immediately. if you reduce timeout, it will not force device to communicate faster, but instead it will fail faster, when device doesn't respond.

xanadu-lee:
When the amount of device data is large, for example, more than 100GB. It may take more than 2 minutes or more to wait from the start of the wait to the first data transfer.
The current mandatory 30-second timeout may cause some inconvenience to the caller.
It would be more flexible if we could enable the caller to customize the timeout.
For example, I perform device synchronization in a thread. Because I need to wait for too long, when I want to cancel before the timeout, I need to wait for the timeout to return before proceeding to the next step.

mexmer:
hmm, now i see what you mean
also disregard my cancel callback comment ... it's nowhere to propagate in libimobile interface, so probly using short timeout call and checking for "timeout error" and do retry, or jump out, is probly best way atm.

mexmer:
just a sidenote, during testing on different phones we found out, that canceling backup (either trough device disconnection, or by sending back error to device, to cancel it), sometimes causes backup service to get stuck in wrong state, and you need to reboot phone, to do backup again.

PR #1401, opened 2023-01-10
Fix iOS 5 SSL error

tihmstar:
Set SSL_OP_IGNORE_UNEXPECTED_EOF on OpenSSL 3 to fix talking to iOS 5

### DIFF
--- a/src/idevice.c
+++ b/src/idevice.c
@@ -1207,13 +1207,26 @@ LIBIMOBILEDEVICE_API idevice_error_t idevice_connection_enable_ssl(idevice_conne
 		SSL_CTX_set_max_proto_version(ssl_ctx, TLS1_VERSION);
 	}
 #endif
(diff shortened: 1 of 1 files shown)


mexmer:
fix checking openssl version
in older versions, there is no OPENSSL_VERSION_MAJOR macro, use
```
#if OPENSSL_VERSION_NUMBER >= 0x030000000
```
instead

nikias:
Turns out that `SSL_CTX_set_options` does *not* clear options that have been set before, so I could solve this even more elegantly: a15a31611d61379297ec45c2a5d34ab57dfb958d

mexmer:
@nikias i would still not use OPENSSL_VERSION_MAJOR macro, it's bad practice to compare undefined macro value, while most preprocessors simple replace undefined macro with empty value, it's just not right

nikias:
@mexmer fair enough: 9a5e78fe56267d6c721c92055c3ae9702bd25871

PR #1432, opened 2023-04-11
src: Update for LibreSSL >= 3.5.0

orbea:
This fixes the build with LibreSSL `>= 3.5.0` and updates it for LibreSSL `>=3.6.0`.

### DIFF
--- a/src/idevice.c
+++ b/src/idevice.c
@@ -1181,7 +1181,8 @@ LIBIMOBILEDEVICE_API idevice_error_t idevice_connection_enable_ssl(idevice_conne
 		return ret;
 	}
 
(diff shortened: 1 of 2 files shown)


nikias:
This won't do. We still need to support the older versions of LibreSSL. Instead of removing stuff we need to add the right version checks if the feature is already available in LibreSSL/OpenSSL.

orbea:
> This won't do. We still need to support the older versions of LibreSSL.

Thanks for the prompt reply, give me some time and I will update this PR.

orbea:
@nikias I updated PR, it was not as much effort as I imagined. I am unsure about the init/deinit code paths and they do not seem to be causing problems so I left them alone.

orbea:
@nikias When you have a chance can you please review this PR?

nikias:
Merged with d87213038d0a428adb3e91432a1b7f6ae4817993.

PR #1442, opened 2023-05-07
Remove bashisms in configure.ac

l-pt:
This change fixes the following errors when building from source on systems where /bin/sh is not bash: 
```
$ ./autogen.sh
[...]
./configure: 20360: test: xno: unexpected operator
./configure: 20372: test: xno: unexpected operator
./configure: 20385: test: xno: unexpected operator
./configure: 20421: test: xyes: unexpected operator
./configure: 20509: test: xno: unexpected operator
configure: error: No SSL library configured. libimobiledevice cannot be built without a supported SSL library.
```


nikias:
Merged with b18bfef.

PR #1444, opened 2023-05-10
include unistd.h for usleep()

kraj:
clang16 flags the missing header

Fixes
../../git/tools/idevicedevmodectl.c:363:2: error: call to undeclared function 'usleep'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]

### DIFF
--- a/tools/idevicedevmodectl.c
+++ b/tools/idevicedevmodectl.c
@@ -40,6 +40,7 @@
 #define __usleep(x) Sleep(x/1000)
 #else
 #include <arpa/inet.h>
+#include <unistd.h>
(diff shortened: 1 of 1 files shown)


mexmer:
usleep and unistd is not available on windows platform if you compiling with msvc. so this fix is not universal.

kraj:
> usleep and unistd is not available on windows platform if you compiling with msvc. so this fix is not universal.

perhaps I should move the include to line 41. Will  upload a new PR

mexmer:
yes it compiles now

nikias:
Merged with d379401.

PR #1510, opened 2023-11-24
Update Linux installation steps

dnicolson:


mexmer:
don't remove libimobiledeviceglue from readme, it is required since move of shared parts in libimobiledevice, libusbmuxd and others to sepratate project https://github.com/libimobiledevice/libimobiledevice-glue
most distribution don't have this package yet in stable branch, because there is no official release, but it's required for succesfull compilation

PR #1532, opened 2024-02-09
find libimobiledevice-glue in e.g. /opt/homebrew/

tallenintegsys:
This allows one to build under OSX using xcode 14.2.

### DIFF
--- a/common/Makefile.am
+++ b/common/Makefile.am
@@ -7,13 +7,15 @@ AM_CFLAGS = \
 	$(ssl_lib_CFLAGS) \
 	$(LFS_CFLAGS) \
 	$(libusbmuxd_CFLAGS) \
-	$(libplist_CFLAGS)
+	$(libplist_CFLAGS) \
+	$(limd_glue_CFLAGS)
 
 AM_LDFLAGS = \
 	$(ssl_lib_LIBS) \
 	${libpthread_LIBS} \
 	$(libusbmuxd_LIBS) \
-	$(libplist_LIBS)
+	$(libplist_LIBS) \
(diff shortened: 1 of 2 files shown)


nikias:
It compiles for me even without the change. The change for `common` makes sense, since it actually uses something from libimobiledevice-glue, but for tools I added the dependency for every tool that uses it. What are you warnings/errors when you try to compile?

tallenintegsys:
Mind you this is on OSX, version 13.6.6...
afcclient.c:72:10: fatal error: ' libimobiledevice-glue/termcolors.h ' file
not found

This has been a while but as I remember I just sent you the changes I had
to make to the makefile.am's (two of them) to get this to compile on macos.
I'm not enough of an auto hell expert to fix this the right way I think I
just assumed you would see the changes that I made and it might be obvious
to you. Presumably something needs to change in the configure (configure.ac)
file. It sounds like you don't have a Mac to compile this on (can't blame
you I detest fake Unix :-) so if you have some suggestions as to what I
should change in the configure file I'll give it a go here and hopefully
just give you a small change to the configure file.

-Tim

tallenintegsys:
Okay, I looked again I think I did make these changes in the correct place.

tallenintegsys:
It's worth noting that while I have been an open source user of a very long time (like when Linux booted off of two floppies) I've not contributed much. As such I don't really know what's expected of me.

tallenintegsys:
This will just be ignored forever? Not sure what else I could have done.

nikias:
This issue was addressed in 469d21c6d506d107a5462c8b10e516f5790c35d3.

PR #1537, opened 2024-02-28
idevicedevmodectl: fix use after free

Fidetro:


nikias:
Not a use-after-free. First idevice_free is not required as creation already failed in the error case.

PR #1539, opened 2024-02-28
Dev

TPJoseph:


jsonlee1688:
Hi @TPJoseph the mobilegestalt has been deprecated for iOS 17.4, will you give a fix for that? if it is, when will you upgrade?

PR #1550, opened 2024-04-06
fix:idevicesetlocation lockdownd handle

Fidetro:


nikias:
See eae0cd26b0aac975737393fb7d87efc871a0f5c2.

PR #1559, opened 2024-04-21
Makefile doesn't compile documentation on first launch

iFred09:
Hello,

When I launch make command for libimobiledevice, it compiles everything but not the docs. I should launch manually ```make docs``` or ```doxygen doxygen.cfg``` after, and the documentation compiles successfully. I precise this information on the README file.


nikias:
Building the docs by default is not really something that everyone needs.

PR #1583, opened 2024-07-19
Fix implicit fallthrough in lockdown.c

r-barnes:
Allows `-Wimplicit-fallthrough` to pass in LLVM.


nikias:
See f1e99c07253e80fe9f1941031d5fb91f329c1a1d.

PR #1591, opened 2024-08-04
Add android exception for pthread detection

JonasVautherin:
I have an issue similar to https://github.com/libimobiledevice/libplist/issues/82 in libimobiledevice. For the same reason, I believe that the pthread detection can be ignored for Android.

This patch fixes my cross-compilation for android-arm64 :+1:.


JonasVautherin:
@nikias any opinion here? 😊

JonasVautherin:
Doesn't seem to be any interest here, closing.

PR #1610, opened 2024-10-17
idevice.c stack corruption

mexmer:
when building on windows for 64bit with openssl version < 3.0 assignment on lines 1084 and 1093 in original code causes stack corruption. this is safe approach, since *processed is not used anyways, but we need value of bytes

stack corruption didn't happen in 32bit build, but i assume, it might also happen, since original construct is really bad.

### DIFF
--- a/src/idevice.c
+++ b/src/idevice.c
@@ -1075,7 +1075,8 @@ static long ssl_idevice_bio_callback(BIO *b, int oper, const char *argp, int arg
 	idevice_connection_t conn = (idevice_connection_t)BIO_get_callback_arg(b);
 #if OPENSSL_VERSION_NUMBER < 0x30000000L
 	size_t len = (size_t)argi;
-	size_t *processed = (size_t*)&bytes;
+	long tempbytes = 0;
+	size_t *processed = (size_t*)&tempbytes;
 #endif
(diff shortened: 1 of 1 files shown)


mexmer:
additional note, compilation was done with MSVC 14.40 (Eg. in VS 2022 LTS version 17.10)

src/idevice.c
```diff
@@ -1075,7 +1075,8 @@ static long ssl_idevice_bio_callback(BIO *b, int oper, const char *argp, int arg
 	idevice_connection_t conn = (idevice_connection_t)BIO_get_callback_arg(b);
 #if OPENSSL_VERSION_NUMBER < 0x30000000L
 	size_t len = (size_t)argi;
-	size_t *processed = (size_t*)&bytes;
+	long tempbytes = 0;
+	long *processed = &tempbytes;
```
nikias, line 1079: I don't really get why the original code would cause stack corruption. Is it about the data type size? `bytes` is aready defined as a `ssize_t`, and the pointer is casted (`size_t*) for the value assignment, and `processed` is also `size_t*`, so how would a compiler mess this up?

mexmer:
i will check that later, but i expect it should not cause issues.

Issue #1355, opened 2022-09-07
SHA-1 getting obsoleted in distributions


hadess:
Fedora, amongst others, is removing SHA-1 support for signatures (eventually) from gnutls and openssl:
https://fedoraproject.org/wiki/Changes/StrongCryptoSettings3Forewarning2

iOS has removed support for SHA-1 TLS certificates, but it obviously still supports that for communication with iOS devices.

So two questions:
- can the SHA-1 signing be transparently replaced with, say, SHA-256 and still support connecting to an iPhone 3GS
- is this signing relevant when connecting to an iDevice over Wi-Fi, as [usbmuxd2 allows](https://github.com/tihmstar/usbmuxd2)?

Downstream bug:
https://bugzilla.redhat.com/show_bug.cgi?id=2124349

hadess:
This change might be needed:
https://github.com/libimobiledevice/libimobiledevice-glue/pull/9
if we need to keep SHA-1 support in libimobiledevice for device support.

mexmer:
you cannot change signing for old devices, they will not understand and communicate with you. it's same issue, why we need to support SSL 3.0, when communicating with ios 1.0 devices. since we can't "patch" old system, to support new ciphers and algorithms, that were not available at time, they were manufactured.

either libimobile will support all apple devices (regardless age), and then you need to support all standards that were and are used, or it will support only "latest greatest", then it can adhere to modern "standards".

from security standpoint, there is no problem with supporting old ciphers and encryption algorithms, libimobile doesn't provide any service, it's just communication bridge, same for ubsmuxd, technically you can connect to it trough socket, but for that, usbmuxd can have communication socket use only latest, while for device communication, it can use all.

but there is issue, with what will be available in distribution, sooner or later you might find out, that libimobile compiled on latest version of your favorite distribution, cannot communicate with 6 years old apple devices, because libraries in your distribution do not have support for old encryption standards ...

hadess:
It's nice of you to answer, but I don't understand how some of it would answer the questions I asked.

> but there is issue, with what will be available in distribution, sooner or later you might find out, that libimobile compiled on latest version of your favorite distribution, cannot communicate with 6 years old apple devices, because libraries in your distribution do not have support for old encryption standards ...

This is also what this issue is about, albeit 

SHA-2 family signatures already existed in 2010, so I'm not sure the comparison with SSL v3 getting deprecated is the right one. 

[...]

PR #1616, opened 2024-11-04
Use SHA256 signature, instead of SHA1 for pairing

mexmer:
it came to my attention that thing discussed in #1355 become issue on more distribution.
so i propose following patch for pairing records creating and signing.

as i mentioned in #1355 , it is not possible to patch old devices to accept SHA256 signed certificates, i have tested it, they will not pair, they will also not do "secure start service".

i had tested this with several devices, for version 4 i have only device with 4.1.2 ... but 4.0.0 was also first iOS support TLS 1.0, (and first iOS with iCloud support), we have similar code in our software since that discussion in #1355 for 2 years without issues.

### DIFF
--- a/common/userpref.c
+++ b/common/userpref.c
@@ -35,7 +35,9 @@
 #ifndef WIN32
 #include <pwd.h>
 #endif
+#ifndef _MSC_VER
 #include <unistd.h>
+#endif
 #include <usbmuxd.h>
 #if defined(HAVE_OPENSSL)
 #include <openssl/bn.h>
@@ -65,7 +67,9 @@
 #endif
 
 #include <dirent.h>
+#ifndef _MSC_VER
 #include <libgen.h>
+#endif
 #include <sys/stat.h>
 #include <errno.h>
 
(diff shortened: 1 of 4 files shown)


common/userpref.c
```diff
@@ -484,7 +488,7 @@ userpref_error_t pair_record_generate_keys_and_certs(plist_t pair_record, key_da
 		X509_set_pubkey(root_cert, root_pkey);
 
 		/* sign root cert with root private key */
-		X509_sign(root_cert, root_pkey, EVP_sha1());
+		X509_sign(root_cert, root_pkey, osversion < DEVICEVERSION(4,0,0) ? EVP_sha1() : EVP_sha256());
```
davide125, line 491: `DEVICEVERSION` doesn't seem to be defined anywhere. `DEVICE_VERSION` _is_ defined in `src/idevice.h`, but that's not included here.

davide125:
I'm backporting this to Fedora in https://src.fedoraproject.org/rpms/libimobiledevice/pull-request/4 and tested it on a few devices. Works fine with iOS 18.3.2 (iPhone 15), iOS 15.8.3 (iPhone 7). With iOS 7.1.2 (iPhone 4s) it connects for second, but disconnects shortly after and the device asks for trust again. This last device is a bit broken, so I don't know how reliable of a test this is.

davide125:
No crypto related errors, I think this is just a faulty device. For the short time it stays connected I can browse it just fine. Going to ship this in Fedora as it's definitely an improvement over the current situation. Will report back  if we get any more useful feedback there.

AsciiWolf:
FWIW, this fix is already used as a downstream patch in distributions such as Fedora 42, Debian Trixie or Arch Linux, and from the user feedback it looks like it works as intended.

nikias:
See c40be2774067703209ec5d3b2a975034b2fb7440

PR #1647, opened 2025-02-27
Fix afcclient

tomriddly:
1. Fix local file name error on pulling folder from device
2. Fix missing last letter of file name when put/get to/from '/'


nikias:
merged after rebase

PR #1661, opened 2025-04-23
Fix LNK2019 for ntohl and ntohs symbol on MSVC

ZeroKwok:


nikias:
See 58baf028aef525e337325c0866b5606f2e3b815c

PR #1662, opened 2025-04-23
When compiled into Unicode and the network address is IPv6, 56 bytes …

xdayong:
…are required here

### DIFF
--- a/src/idevice.c
+++ b/src/idevice.c
@@ -559,7 +559,7 @@ idevice_error_t idevice_connect(idevice_t device, uint16_t port, idevice_connect
 				return IDEVICE_E_UNKNOWN_ERROR;
 		}
 
-		char addrtxt[48];
+		char addrtxt[56];
 		addrtxt[0] = '\0';
 
 		if (!socket_addr_to_string(saddr, addrtxt, sizeof(addrtxt))) {


mexmer:
i highly doubt it.
full size IPv6 address in text form needs 40 byte buffer
it's 8 group with 4 hex numbers, 7 separators, and null terminator
if each character was represented in unicode (UCS2), you will need double length buffer, eg. 80 bytes (or 40 wchars)
your 56 bytes doesn't make sense.

FYI IPv6 full length address looks like this
ffff:ffff:ffff:ffff:ffff:ffff:ffff:ffff:ffff
specification allows for shortening
so instead of 
1234:5678:90ab:cdef:0001:0002:0003:0004 you can write 1234:5678:90ab:cdef:1:2:3:4
also another shortening is skipping 0 value segments so instead of 
1234:5678:90ab:cdef:0000:0000:0000:0001 you can write 1234:5678:90ab:cdef::1

mexmer:
just to add to this. on windows you can call one of following implementations
WSAAddressToString - macro that switches to one of bellow function, based on UNICODE switch for compiler
WSAAddressToStringA - returns address in chars
WSAAddressToStringW - returns address in widechars (UNICODE switch)
see https://learn.microsoft.com/en-us/windows/win32/api/winsock2/nf-winsock2-wsaaddresstostring
on other systems you call
inet_ntop - this function expectes buffer that is INET6_ADDRSTRLEN characters long (which is 46), see https://man7.org/linux/man-pages/man3/inet_ntop.3.html

xdayong:
> i highly doubt it. full size IPv6 address in text form needs 40 byte buffer it's 8 group with 4 hex numbers, 7 separators, and null terminator if each character was represented in unicode (UCS2), you will need double length buffer, eg. 80 bytes (or 40 wchars) your 56 bytes doesn't make sense.
> 
> FYI IPv6 full length address looks like this ffff:ffff:ffff:ffff:ffff:ffff:ffff:ffff:ffff specification allows for shortening so instead of 1234:5678:90ab:cdef:0001:0002:0003:0004 you can write 1234:5678:90ab:cdef:1:2:3:4 also another shortening is skipping 0 value segments so instead of 1234:5678:90ab:cdef:0000:0000:0000:0001 you can write 1234:5678:90ab:cdef::1

Oh, You are right, thank you for your explanation. But here, I suggest changing the buf size to 80, so that when compiling Unicode, other parts do not need to be modified.

mexmer:
> > i highly doubt it. full size IPv6 address in text form needs 40 byte buffer it's 8 group with 4 hex numbers, 7 separators, and null terminator if each character was represented in unicode (UCS2), you will need double length buffer, eg. 80 bytes (or 40 wchars) your 56 bytes doesn't make sense.
> > FYI IPv6 full length address looks like this ffff:ffff:ffff:ffff:ffff:ffff:ffff:ffff:ffff specification allows for shortening so instead of 1234:5678:90ab:cdef:0001:0002:0003:0004 you can write 1234:5678:90ab:cdef:1:2:3:4 also another shortening is skipping 0 value segments so instead of 1234:5678:90ab:cdef:0000:0000:0000:0001 you can write 1234:5678:90ab:cdef::1
> 
> Oh, You are right, thank you for your explanation. But here, I suggest changing the buf size to 80, so that when compiling Unicode, other parts do not need to be modified.

technically, this is required only for windows compilation, due to use of WSA function, other option will be to use ansi version of function even when compiled for unicode, eg. calling WSAAddressToStringA instead of WSAAddressToString, that changes depending on unicode switch. since function has defined only char version
```
LIMD_GLUE_API const char *socket_addr_to_string(struct sockaddr *addr, char *addr_out, size_t addr_out_size);
```
https://github.com/libimobiledevice/libimobiledevice-glue/blob/c915351cb322d041afabc04f780eb35142cdaea5/include/libimobiledevice-glue/socket.h#L72C1-L72C110

PR #1671, opened 2025-06-07
Add unmatch filter option to idevicesyslog.

xingheng:
The new -M/--unmatch option allows filtering out messages containing specific strings, complementing the existing -m/--match functionality. This provides more flexible log filtering capabilities for users.


nikias:
See 4c50979bdf20d8d17281463cef7cd4c848f6d79d

PR #1697, opened 2025-11-02
libsrp6a-sha512: fix environ for Apple

barracuda156:
Fix `environ` declaration


nikias:
Hi, I tried both code paths on macOS and both yield the same results. Please clarify why this change would make a difference?

barracuda156:
Thank you for merging!

PR #1716, opened 2026-02-03
Support for root mode / afc2

rickmark:
Adds basic support for `com.apple.afc2` (AFC while in root mode)


rickmark:
@nikias  - any reason for the stall on this PR?

PR #1748, opened 2026-08-17
idevicebackup: Open each received file once instead of per hunk

kaolin:
### Problem

`idevicebackup` (the v1 `com.apple.mobilebackup` protocol, used by iOS 3.x and
earlier) silently corrupts every file larger than one hunk.

File data arrives from the device in hunks. The receive loop calls
`buffer_write_to_filename()` once per hunk — and that helper opens the
destination with `"wb"`, truncating it every time. Each hunk therefore
overwrites the previous one and only the **last** hunk survives.

The failure is silent: the tool prints `Backup Successful.`, the manifest is
complete, and every file is present. Only the contents are wrong.

Observed against an iPhone 3G (`iPhone1,2`, iOS 3.1):

| file | actual size | size in backup |
|---|---|---|
| `IMG_0176.JPG` | 331,115 B | 3,435 B |
| `sms.db` | 200,704 B | 4,096 B |
| `AddressBook.sqlitedb` | 237,568 B | 8,192 B |
| `notes.db` | 45,056 B | 4,096 B |

Every SQLite database arrived as a single orphaned b-tree page with no
`SQLite format 3` header, so nothing in the backup could be opened at all.
The recovered 3,435 bytes of `IMG_0176.JPG` are byte-for-byte the tail of the
real file, starting at offset 327,680 — exactly 10 × 32 KB.

### The intent was already append

Three lines above the bug:

```c
/* if this is the first hunk, remove any existing file */
if ((hunk_index == 0) && (stat(filename_mddata, &st) == 0))
    remove(filename_mddata);
```

Removing the stale file only on hunk 0 is only meaningful if the writes that
follow append. The helper that was reached for truncates instead.

### Fix

`libimobiledevice-glue` exposes no append variant, so this adds a local
`buffer_append_to_filename()` and uses it for the data hunks.

### Verification

Same device, same backup, before and after:

- backup size: **13 MB → 74 MB** (1319 files both times)
- all six `HomeDomain` databases now carry a valid `SQLite format 3` header and
  open cleanly (13 notes, 799 SMS, 103 contacts, 100 calls, 11 voicemails)
- a photo extracted from the backup is now **byte-identical** to the same file
  pulled independently over AFC — matching SHA1 `8ed1ef9c724f…`

That last check is the real proof: two unrelated transports agreeing exactly.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


mexmer:
instead of fixing single function like that. would be better to rework whole receive file, to open/create file upon receiving first hunk, then write all received hunks, and close it afterwards.

apple backup doesn't leave hunks from different files.
filename and attributes are received only with hunk 0, other hunks don't carry filename, just data.

kaolin:
Good call — done. Rewrote it as a proper open/write/close rather than patching the single write, and dropped the `buffer_append_to_filename` helper entirely.

The destination is resolved once from `DLFileDest` on hunk 0, the file is opened there, each hunk written to the open handle, and closed on `DEVICE_LINK_FILE_STATUS_LAST_HUNK`. The manifest `rename()` moved after the close, and the paths that leave the loop mid-file close whatever is still open.

Your point about attributes only arriving with hunk 0 mattered more than it first looked: the old code called `plist_get_string_val(node, &file_path)` on *every* hunk, and since `DLFileDest` isn't in hunks 1..n, `file_path` kept the value from hunk 0 — it was working off a stale pointer, and leaking a rebuilt path per hunk. Fetching it once makes that explicit. It also means `file_path` can now legitimately be NULL, and `mobilebackup_build_path()` does `strlen(name)` unguarded, so there's a check — a missing filename sets `quit_flag` and takes the existing cancel path rather than segfaulting. Failed `fopen`/`fwrite` do the same.

Tested against an iPhone 3G (iPhone1,2) on iOS 3.1, full backup, 1319 files. Before: 13 MB of truncated fragments. After: 68 MB, matching what the device reports. I still had the tree from the earlier append-based build, so I diffed the two backups file by file — 2634 of 2638 `.mddata`/`.mdinfo` are byte-identical, and the 4 that differ are live state that changed between runs (`com.apple.wifi.plist`, `com.apple.springboard.plist`, `com.apple.AutoWake.plist`, `ADDataStore.sqlitedb`), all with intact headers. Every photo and every user database is byte-identical, and files extracted from the backup match the same files pulled independently over AFC.

PR #1752, opened 2026-08-18
build: discover python3 before Python development check

### LINKED ISSUE #1706: [BUG] `./autogen.sh` fails if `python` isn't found, even when `python3` is installed.
Trying to build on an odd system, a WiiU (see https://linux-wiiu.org), using Adelie Linux (see https://www.adelielinux.org/). For some reason, it doesn't try to check for `python3` in `$PATH` when it probably should.

Please let me know if more information is needed.

I'm using the latest commit as of 12/22/25 (commit [c4f1118](https://github.com/libimobiledevice/libimobiledevice/commit/c4f111800d2b08d6c65ec075ad49c4a60e9e4309))

tmsteph:
`CYTHON_PYTHON` currently uses `AC_REQUIRE([AX_PYTHON_DEVEL])`. Autoconf therefore emits the development check before the surrounding `AM_PATH_PYTHON([3.0], ...)` call, so systems that provide `python3` but no `python` fail before Automake gets a chance to discover `python3`.

Invoke `AX_PYTHON_DEVEL` in place instead. This lets `AM_PATH_PYTHON` select a Python 3 interpreter first; `AX_PYTHON_DEVEL` then reuses that `PYTHON` value for the development checks.

Validation:
- regenerated `configure` with `NOCONFIGURE=1 ./autogen.sh`
- confirmed the Automake `python python3 ...` search now precedes the AX_PYTHON_DEVEL version/development checks
- ran `configure` with `PATH=/usr/bin:/bin` on a host where `python3` is available; verified the old `Cannot find python...` failure is absent
- `git diff --check`

Fixes #1706.

### DIFF
--- a/m4/cython_python.m4
+++ b/m4/cython_python.m4
@@ -1,6 +1,6 @@
 AC_DEFUN([CYTHON_PYTHON],[
         AC_REQUIRE([AC_PROG_CYTHON])
-        AC_REQUIRE([AX_PYTHON_DEVEL])
+        AX_PYTHON_DEVEL
         test "x$1" != "xno" || cython_shadow=" -noproxy"
         AC_SUBST([CYTHON_PYTHON_OPT],[-python$cython_shadow])
         AC_SUBST([CYTHON_PYTHON_CPPFLAGS],[$PYTHON_CPPFLAGS])


tmsteph:
Rechecked this against current master: m4/cython_python.m4 still has AC_REQUIRE([AX_PYTHON_DEVEL]), so the fix remains applicable and the PR is still cleanly mergeable. The recorded Actions run currently reports failure but has zero jobs, so there is no actionable test failure from the patch itself to address. Happy to adjust if maintainers prefer a different Autoconf ordering approach.

